PR #10, opened 2018-11-14
[CEDS - 740]- save and get notifications

raghu1978:


app/uk/gov/hmrc/exports/models/Submission.scala
```diff
@@ -27,15 +27,12 @@ case class Submission(
   mrn: Option[String] = None,
   submittedTimestamp: Long = System.currentTimeMillis(),
   id: BSONObjectID = BSONObjectID.generate(),
-  arrivalConversationId: Option[String] = None,
-  departureConversationId: Option[String] = None
+  status: Option[String] = Some("Pending")
```
PatrykRudnicki: Propose for the future trait for status

PR #20, opened 2019-01-29
CEDS-930 Update status of declaration submission

PatrykRudnicki:


app/uk/gov/hmrc/exports/controllers/ExportController.scala
```diff
@@ -31,20 +31,17 @@ import scala.concurrent.{ExecutionContext, Future}
...
-  def authorizedWithEnrolment[A](
-    callback: (Request[A]) => Future[Result]
+  def authorizedWithEori[A](
+    callback: (AuthorizedRequest[A] => Future[Result])
   )(implicit request: Request[A]): Future[Result] =
     authorised(Enrolment("HMRC-CUS-ORG")).retrieve(allEnrolments) { enrolments =>
-      if (!hasEnrolment(enrolments))
-        callback(request)
+      val eori = enrolments.getEnrolment("HMRC-CUS-ORG").flatMap(_.getIdentifier("EORINumber"))
+      if (!eori.isEmpty) callback(AuthorizedRequest(request, eori.get.value))
```
MaciejRewera: eori.nonEmpty would be better.
And to increase readibility I would suggest wrapping the line no. 38 into private method as it used to be. Maybe call it sth like getEoriFrom(enrolments: ...)

PR #23, opened 2019-02-21
CEDS-939 Notification status

PatrykRudnicki:


app/uk/gov/hmrc/exports/controllers/SubmissionController.scala
```diff
@@ -128,10 +128,7 @@ class SubmissionController @Inject()(
     submissionRepository.findByEori(eori)
 
   private def notificationHelper(conversationId: String): Future[Int] =
```
MaciejRewera: This method name is completely non-explanatory. Could you change the name to reflect what's being done?

PatrykRudnicki: Yes, with next PR with logic

PR #32, opened 2019-03-25
CEDS-1062: refactored the header validation in notification controlle…

mattclark-zerogravit:
…r and added additional logging for error cases

### DIFF
--- a/app/uk/gov/hmrc/exports/controllers/HeaderValidator.scala
+++ b/app/uk/gov/hmrc/exports/controllers/HeaderValidator.scala
@@ -17,25 +17,49 @@
 package uk.gov.hmrc.exports.controllers
 
 import javax.inject.Singleton
+import play.api.Logger
+import play.api.http.HeaderNames
 import uk.gov.hmrc.exports.models._
+
 @Singleton
 class HeaderValidator {
 
   def extractLrnHeader(headers: Map[String, String]): Option[String] = {
-    headers.get(CustomsHeaderNames.XLrnHeaderName)
+    extractHeader(CustomsHeaderNames.XLrnHeaderName, headers)
   }
 
   def extractDucrHeader(headers: Map[String, String]): Option[String] = {
-    headers.get(CustomsHeaderNames.XDucrHeaderName)
+    extractHeader(CustomsHeaderNames.XDucrHeaderName, headers)
   }
 
   def extractMrnHeader(headers: Map[String, String]): Option[String] = {
-    headers.get(CustomsHeaderNames.XMrnHeaderName)
+    extractHeader(CustomsHeaderNames.XMrnHeaderName, headers)
   }
 
+  def extractAuthTokenHeader(headers: Map[String, String]): Option[String] = {
+    extractHeader(HeaderNames.AUTHORIZATION, headers)
+  }
+
+  def extractConversationIdHeader(
+                                   headers: Map[String, String]): Option[String] = {
+    extractHeader(CustomsHeaderNames.XConversationIdName, headers)
+  }
+
+  def extractEoriHeader(headers: Map[String, String]): Option[String] = {
+    extractHeader(CustomsHeaderNames.XEoriIdentifierHeaderName, headers)
+  }
+
+  private def extractHeader(headerName: String,
+                            headers: Map[String, String]): Option[String] =
+    headers.get(headerName) match {
+      case Some(header) if !header.isEmpty => Some(header)
+      case _ =>
+        Logger.error(s"Error Extracting $headerName")
+        None
+    }
 
   def validateAndExtractSubmissionHeaders(implicit headers: Map[String, String])
-    : Either[ErrorResponse, ValidatedHeadersSubmissionRequest] = {
+  : Either[ErrorResponse, ValidatedHeadersSubmissionRequest] = {
     val result = for {
       lrn <- extractLrnHeader(headers)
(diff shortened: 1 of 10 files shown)


app/uk/gov/hmrc/exports/controllers/HeaderValidator.scala
```diff
@@ -17,22 +17,47 @@
...
+    extractHeader(CustomsHeaderNames.XConversationIdName, headers)
+  }
+
+  def extractEoriHeader(headers: Map[String, String]): Option[String] = {
+    headers.get("X-EORI-Identifier")
+  }
+
+  private def extractHeader(headerName: String,
+                            headers: Map[String, String]): Option[String] = {
+    headers.get(headerName) match {
```
PatrykRudnicki: I don't like this code. Why?
1) if option is defined return the same without any changes
2) if option is empty return None with Logger.

Here we need to only log a situation when the Option is not empty, we don't need to check anything else. I checked Option documentation and I found method that will be good for this:

https://www.scala-lang.org/api/current/scala/Option.html#orElse[B%3E:A](alternative:=%3EOption[B]):Option[B]

I checked documentation for version that we use and this method exists.

```
headers.get(headerName).orElse{
    Logger.error(s"Error Extracting $headerName")
    None
```

This solution is not probably the best one, but it's works and probably it should be faster:

```
/** Returns this $option if it is nonempty,
   *  otherwise return the result of evaluating `alternative`.
   *  @param alternative the alternative expression.
   */
  @inline final def orElse[B >: A](alternative: => Option[B]): Option[B] =
      if (isEmpty) alternative else this
```

I'm thinking there should be better solution, but I can't remember which one.

app/uk/gov/hmrc/exports/controllers/HeaderValidator.scala
```diff
@@ -44,22 +69,44 @@ class HeaderValidator {
     result match {
       case Some(vhr) =>
         Right(vhr)
-      case _ =>
-        Left(ErrorResponse.ErrorInternalServerError)
+      case None => {
```
PatrykRudnicki: Pattern matching doesn't need to have parenthesis. You can have it like:
```
case None =>
    Logger.debug(....)
    Left(.....)
```

app/uk/gov/hmrc/exports/controllers/NotificationsController.scala
```diff
@@ -54,7 +56,19 @@ class NotificationsController @Inject()(
   def saveNotification(): Action[NodeSeq] = Action.async(parse.xml) {
     implicit request =>
       metrics.startTimer(notificationMetric)
-      validateHeaders(getNotificationFromRequest _ andThen save _)
+      implicit val body: NodeSeq = request.body
```
PatrykRudnicki: Where you need this? I'm not sure but you can remove it

app/uk/gov/hmrc/exports/controllers/SubmissionController.scala
```diff
@@ -121,21 +121,21 @@ class SubmissionController @Inject()(
...
-              case e: Exception =>
-                Logger.error(s"problem calling declaration api ${e.getMessage}")
-                Future.successful(
-                  ErrorResponse.ErrorInternalServerError.XmlResult)
-            }
+            handleDeclarationCancelation(request.eori.value, vhr.mrn.value, xml)
+              .recoverWith {
+                case e: Exception =>
+                  Logger.error(
+                    s"problem calling declaration api ${e.getMessage}")
```
PatrykRudnicki: Please start log with capital letter

app/uk/gov/hmrc/exports/controllers/HeaderValidator.scala
mattclark-zerogravit: I've added a guard so it will only return some when there is a value.

app/uk/gov/hmrc/exports/controllers/SubmissionController.scala
mattclark-zerogravit: done

app/uk/gov/hmrc/exports/controllers/HeaderValidator.scala
```diff
@@ -44,22 +69,44 @@ class HeaderValidator {
...
+        Logger.debug("error validating and extracting headers")
```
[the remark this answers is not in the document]
mattclark-zerogravit: done

app/uk/gov/hmrc/exports/controllers/HeaderValidator.scala
```diff
@@ -17,22 +17,47 @@
...
+    extractHeader(HeaderNames.AUTHORIZATION, headers)
+  }
+
+  def extractConversationIdHeader(
+      headers: Map[String, String]): Option[String] = {
+    extractHeader(CustomsHeaderNames.XConversationIdName, headers)
+  }
+
+  def extractEoriHeader(headers: Map[String, String]): Option[String] = {
+    headers.get("X-EORI-Identifier")
```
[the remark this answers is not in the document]
mattclark-zerogravit: done

app/uk/gov/hmrc/exports/controllers/NotificationsController.scala
mattclark-zerogravit: yep will do

PR #33, opened 2019-03-26
CEDS-1071: Made the ducr optional in the model and as a header

mattclark-zerogravit:


app/uk/gov/hmrc/exports/controllers/HeaderValidator.scala
```diff
@@ -27,8 +27,8 @@ class HeaderValidator {
   def extractLrnHeader(headers: Map[String, String]): Option[String] =
     extractHeader(CustomsHeaderNames.XLrnHeaderName, headers)
 
-  def extractDucrHeader(headers: Map[String, String]): Option[String] =
-    extractHeader(CustomsHeaderNames.XDucrHeaderName, headers)
+  def extractOptionalDucrHeader(headers: Map[String, String]): Option[String] =
+    headers.get(CustomsHeaderNames.XDucrHeaderName)
```
PatrykRudnicki: Why are you come back to version with `headers.get` instead of `extractHeader`?

mattclark-zerogravit: As its optional header so no need to print out error in log file if it's missing.

PR #34, opened 2019-03-27
CEDS-1058: mrn is now parsed from the notification and have updated h…

mattclark-zerogravit:
…andling of headers as eori will never be received for a notification

### DIFF
--- a/app/uk/gov/hmrc/exports/repositories/SubmissionRepository.scala
+++ b/app/uk/gov/hmrc/exports/repositories/SubmissionRepository.scala
@@ -70,10 +70,10 @@ class SubmissionRepository @Inject()(implicit mc: ReactiveMongoComponent, ec: Ex
     }
   }
 
-  def updateStatus(eori: String, convId: String, status: Option[String]): Future[Boolean] =
+  def updateMrnAndStatus(eori: String, convId: String, mrn: String, status: Option[String]): Future[Boolean] =
     if (status.isDefined) {
       val finder = BSONDocument("eori" -> eori, "conversationId" -> convId)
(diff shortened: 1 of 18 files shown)


app/uk/gov/hmrc/exports/repositories/SubmissionRepository.scala
```diff
@@ -70,10 +70,10 @@ class SubmissionRepository @Inject()(implicit mc: ReactiveMongoComponent, ec: Ex
     }
   }
 
-  def updateStatus(eori: String, convId: String, status: Option[String]): Future[Boolean] =
+  def updateStatus(eori: String, convId: String, mrn: String, status: Option[String]): Future[Boolean] =
     if (status.isDefined) {
       val finder = BSONDocument("eori" -> eori, "conversationId" -> convId)
-      val modifier = BSONDocument("$set" -> BSONDocument("status" -> status.get))
+      val modifier = BSONDocument("$set" -> BSONDocument("mrn" -> mrn, "status" -> status.get))
```
PatrykRudnicki: Hmm, I think that for this case we shouldn't update `mrn` and `status` inside `updateStatus` method. This method should only update status. You can use diffrent one `updateSubmission`, for me it's not a good place to do it.

PatrykRudnicki:
I know that we don't have the best style configuration, but it'll be good to have consistent code. Could you run scalafmt?

app/uk/gov/hmrc/exports/repositories/SubmissionRepository.scala
mattclark-zerogravit: I've renamed the method updateStatusAndMrn. Its only called in one place.
Scalafmt run as well. but changed loads of files I've not modified..

PR #36, opened 2019-03-28
CEDS-1074: fixed MRN parsing and tightened up the tests to cover real…

mattclark-zerogravit:
… example of xml response

MRN was being parsed from the incorrect declaration in Metdata. We need to use the one inside response.declaration

### DIFF
--- a/app/uk/gov/hmrc/exports/controllers/NotificationsController.scala
+++ b/app/uk/gov/hmrc/exports/controllers/NotificationsController.scala
@@ -109,7 +109,7 @@ class NotificationsController @Inject()(
 
     parseXmlResult match {
       case Success(metaData) =>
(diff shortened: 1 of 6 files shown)


app/uk/gov/hmrc/exports/controllers/NotificationsController.scala
```diff
@@ -109,7 +109,7 @@ class NotificationsController @Inject()(
 
     parseXmlResult match {
       case Success(metaData) =>
-        val mrn = metaData.declaration.flatMap(_.functionalReferenceId)
+        val mrn = metaData.response.head.declaration.flatMap(_.id)
```
PatrykRudnicki: Please don't use `head`. Use `headOption` it'll be safer, when this field will be empty

PR #39, opened 2019-04-29
[CEDS-662- - metrics update and formatting

raghu1978:


test/uk/gov/hmrc/exports/base/CustomsExportsBaseSpec.scala
```diff
@@ -74,6 +70,8 @@ trait CustomsExportsBaseSpec
 
   def appConfig: AppConfig = injector.instanceOf[AppConfig]
 
+  val metrics: ExportsMetrics = injector.instanceOf[ExportsMetrics]
```
PatrykRudnicki: Why this? I think using mock for tests instead of inject is better

raghu1978: I want to test the metrics so have injected the original

PatrykRudnicki: 👍 Approved

PR #59, opened 2019-08-06
CEDS-1443 Update submission if notification came earlier 

bambuchaAdm:


test/unit/uk/gov/hmrc/exports/base/CustomsExportsBaseSpec.scala
```diff
@@ -127,10 +128,13 @@ trait CustomsExportsBaseSpec
   protected def withSubmissions(submissions: Seq[Submission]): OngoingStubbing[Future[Seq[Submission]]] =
     when(mockSubmissionRepository.findAllSubmissionsForEori(any())).thenReturn(Future.successful(submissions))
 
-//  protected def withNotification(
-//    notifications: Seq[Notification]
-//  ): OngoingStubbing[Future[Seq[Notification]]] =
-//    when(mockNotificationsRepository.getByConversationId(any())).thenReturn(Future.successful(notifications))
+  protected def withoutNotifications(): OngoingStubbing[Future[Seq[Notification]]] =
```
MaciejRewera: Just for the record: 
I think we should not keep adding more methods to the base spec. We had enough situations where it grew too big to maintain. IMHO it is better to repeat yourself in tests, or create a method for the suite scope only.
And in this case the method name is not fully explanatory. I'd suggest sth like `withoutNotificationsInDB` and `withNotificationsInDB`.

bambuchaAdm: I have putted this method exactly where mock lives. This type of refactor would require bigger changes in this service i think is out of scope for this ticket. 

I fully agree on this movement - big base class cause most of problems - specially when it starts classes.

PR #60, opened 2019-08-08
CEDS-1479 Repository Layer

ed0906:


app/uk/gov/hmrc/exports/models/declaration/ExportsDeclaration.scala
```diff
@@ -0,0 +1,30 @@
...
+import play.api.libs.json.{JsError, JsNumber, JsObject, JsResult, JsSuccess, JsValue, Json, OFormat}
+
+case class ExportsDeclaration
+(
+  id: String = UUID.randomUUID().toString,
+  eori: String
+)
+
+object ExportsDeclaration {
+  implicit val formatInstant: OFormat[LocalDateTime] = new OFormat[LocalDateTime] {
```
ed0906: This is required to save dates in the format Mongo wants rather than as a String

ed0906: In hindsight It should be an Instant as agreed in the last tech meeting. Will change in subsequent PRs

PatrykRudnicki: so we're storing dates in mongo in EpochMilli format?

PR #61, opened 2019-08-08
CEDS-1479 Create Endpoint to persist to the backend

ed0906:
A lightweight endpoint to Persist `ExportsDeclaration`s to the back end

### DIFF
--- a/app/uk/gov/hmrc/exports/controllers/DeclarationController.scala
+++ b/app/uk/gov/hmrc/exports/controllers/DeclarationController.scala
@@ -19,6 +19,7 @@ package uk.gov.hmrc.exports.controllers
 import javax.inject.{Inject, Singleton}
 import play.api.libs.json.Json
 import play.api.mvc._
+import uk.gov.hmrc.exports.controllers.actions.Authenticator
 import uk.gov.hmrc.exports.models.declaration.ExportsDeclaration
 import uk.gov.hmrc.exports.services.DeclarationService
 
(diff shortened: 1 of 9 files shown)


app/uk/gov/hmrc/exports/controllers/DeclarationController.scala
```diff
@@ -0,0 +1,39 @@
...
+
+import scala.concurrent.ExecutionContext.Implicits.global
+
+@Singleton
+class DeclarationController @Inject()(
+  declarationService: DeclarationService,
+  override val controllerComponents: ControllerComponents
+) extends RESTController {
+
+  def post(): Action[ExportsDeclaration] = Action.async(parsingJson[ExportsDeclaration]) { implicit request =>
```
ed0906: Do we think this needs any kind of Auth?

ed0906:
I've done this as a completely new slice (new Controller+Service+Repository) as I suspect a lot of the old stuff will end up getting replaced

app/uk/gov/hmrc/exports/controllers/DeclarationController.scala
bambuchaAdm: I think so - it even describe in platform patterns - [Assumed Upstream Security](https://confluence.tools.tax.service.gov.uk/display/AR/Assumed+Upstream+Security)

ed0906: Auth Added

PR #64, opened 2019-08-09
CEDS-1479 Populate more of the model

ed0906:


app/uk/gov/hmrc/exports/models/declaration/DUCR.scala
```diff
@@ -0,0 +1,25 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package uk.gov.hmrc.exports.models.declaration
+
+import play.api.libs.json.{Json, OFormat}
+
+case class DUCR(ducr: String)
```
bambuchaAdm: this is classical value class - maybe this field should name `value` ?

app/uk/gov/hmrc/exports/models/declaration/DUCR.scala
```diff
@@ -0,0 +1,25 @@
...
+
+object DUCR {
+  implicit val format: OFormat[DUCR] = Json.format[DUCR]
```
bambuchaAdm: this would lead to something like 
```
{
  "ducr" : {
     "ducr" : "value"
  }
}
```
in ConsignmentReferences - i really do not know do w really want to have something like this.

app/uk/gov/hmrc/exports/models/declaration/DeclarationStatus.scala
```diff
@@ -0,0 +1,26 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package uk.gov.hmrc.exports.models.declaration
+
+import play.api.libs.json.Format
+import uk.gov.hmrc.exports.util.EnumJson
+
+object DeclarationStatus extends Enumeration {
```
bambuchaAdm: Personally i don't like scala `Enumeration` - sealed traits works better IMO

PatrykRudnicki: Agree, but reads/writes always are more complicated with sealed traits

PR #66, opened 2019-08-12
CEDS-1481 Retrieve Declarations by ID & EORI

ed0906:


app/uk/gov/hmrc/exports/services/DeclarationService.scala
```diff
@@ -30,4 +30,8 @@ class DeclarationService @Inject()(declarationRepository: DeclarationRepository)
    */
   def save(declaration: ExportsDeclaration): Future[ExportsDeclaration] = declarationRepository.create(declaration)
 
+  def find(eori: String): Future[Seq[ExportsDeclaration]] = declarationRepository.find(eori)
```
bambuchaAdm: I am not sure which declaration would be returned when there is more then one submitted.

bambuchaAdm: Ok there is seq there so is should not be problem.

conf/app.routes
```diff
@@ -1,6 +1,8 @@
 # microservice specific routes
 
 POST        /v2/declaration                                  uk.gov.hmrc.exports.controllers.DeclarationController.post()
+GET         /v2/declaration                                  uk.gov.hmrc.exports.controllers.DeclarationController.get()
```
PauloPortugal: :+1:

test/unit/uk/gov/hmrc/exports/controllers/DeclarationControllerSpec.scala
```diff
@@ -56,14 +56,14 @@ class DeclarationControllerSpec
     reset(mockAuthConnector, declarationService)
   }
 
-  "POST" should {
+  "POST /" should {
     val post = FakeRequest("POST", "/v2/declaration")
 
-    "return 200" when {
+    "return 201" when {
```
PauloPortugal: :+1: Awesome!

app/uk/gov/hmrc/exports/services/DeclarationService.scala
ed0906: Yeh I eventually see this superseding the "get by eori: endpoint

PR #67, opened 2019-08-12
CEDS-1500 Remove Declaration by EORI and ID

ed0906:


app/uk/gov/hmrc/exports/controllers/DeclarationController.scala
```diff
@@ -48,7 +51,16 @@ class DeclarationController @Inject()(
...
     declarationService.findOne(id, request.eori.value).map {
       case Some(declaration) => Ok(Json.toJson(declaration))
-      case None => NotFound
+      case None              => NotFound
+    }
+  }
+
+  def deleteByID(id: String): Action[AnyContent] = authenticator.authorisedAction(parse.default) { implicit request =>
+    declarationService.findOne(id, request.eori.value).flatMap {
+      case Some(declaration) if declaration.status != DeclarationStatus.COMPLETE =>
```
MaciejRewera: I would move some of this logic into Service layer. It is not something that Controller needs to care about.

conf/app.routes
```diff
@@ -3,6 +3,7 @@
 POST        /v2/declaration                                  uk.gov.hmrc.exports.controllers.DeclarationController.post()
 GET         /v2/declaration                                  uk.gov.hmrc.exports.controllers.DeclarationController.get()
 GET         /v2/declaration/:id                              uk.gov.hmrc.exports.controllers.DeclarationController.getByID(id: String)
+DELETE      /v2/declaration/:id                              uk.gov.hmrc.exports.controllers.DeclarationController.deleteByID(id: String)
```
MaciejRewera: I know that it is a bit late now, but would it not be more readable if we use more explanatory method names? `post` or `get` are not very descriptive, since they only tell about request types.

app/uk/gov/hmrc/exports/controllers/DeclarationController.scala
bambuchaAdm: It should be even on repository level

PR #71, opened 2019-08-13
CEDS-1478 Ensure REST JSON Formatter is used in controllers

ed0906:


bambuchaAdm:
Maybe we should just use explicit passing those formats (reads/writes) ?

app/uk/gov/hmrc/exports/models/declaration/ExportsDeclaration.scala
```diff
@@ -44,19 +44,26 @@ case class ExportsDeclaration(
...
-            case _           => JsError("Unexpected Date Format. Expected a Number (Epoch Milliseconds)")
-          }
-        case _ => JsError("Unexpected Date Format. Expected an object containing a $date field.")
-      }
+  object REST {
+    implicit val format: OFormat[ExportsDeclaration] = Json.format[ExportsDeclaration]
   }
-  implicit val format: OFormat[ExportsDeclaration] = Json.format[ExportsDeclaration]
+
+  object Mongo {
```
PauloPortugal: Why do we need to create a `Mongo` and `REST` object?
Is it just to separate the `write` and `read` concerns?

As far as naming goes:
One refers to an architecture style `REST`, the other to a NoSQL tool.
Probably it would make sense to think them as layers: `Presentation` Layer vs a `DataAccess` Layer, but naming is always difficult.

ed0906: The JSON that it serializes to is subtly different, e.g. Mongo stored dates differently.

I've named the object `Mongo` as the format is very specific to that type of DB, If we swapped it out for a different NoSQL DB I'd imagine the format would be completely differentt

ed0906:
> Maybe we should just use explicit passing those formats (reads/writes) ?
> 
> It would just require to pass `ExportsDeclaration.mongo` or `ExportsDeclaration.rest` in few places.

I'm doing the same with imports in the controller layer + repository

bambuchaAdm:
I just don't like when there is bunch of magical imports with implicits - it is easy to mess up with them even through IDE `optimize imports`

PR #72, opened 2019-08-13
CEDS-1478 Add Pagination & Search by Status to Find All Decs by EORI

ed0906:


app/uk/gov/hmrc/exports/models/DeclarationSearch.scala
```diff
@@ -0,0 +1,27 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package uk.gov.hmrc.exports.models
+
+import play.api.libs.json.{Json, OFormat}
+import uk.gov.hmrc.exports.models.declaration.DeclarationStatus.DeclarationStatus
+
+case class DeclarationSearch(eori: String, status: Option[DeclarationStatus] = None)
```
ed0906: From experience we usually end up adding more and more search filters over time so I've stuck it in a Search class for now

app/uk/gov/hmrc/exports/models/Page.scala
```diff
@@ -0,0 +1,45 @@
...
+package uk.gov.hmrc.exports.models
+
+import play.api.libs.json.{Json, OFormat}
+import play.api.mvc.QueryStringBindable
+
+case class Page(index: Int = 1, size: Int = 50)
+
+object Page {
+  implicit val format: OFormat[Page] = Json.format[Page]
+  implicit def bindable(implicit intBinder: QueryStringBindable[Int]): QueryStringBindable[Page] =
```
ed0906: If you haven't seen these before it means we can use complex types in `app.routes` and  Controller params and it "just works"

I expect Notifications will want this eventually too

app/uk/gov/hmrc/exports/models/Paginated.scala
```diff
@@ -0,0 +1,33 @@
...
+package uk.gov.hmrc.exports.models
+
+import play.api.libs.json._
+
+case class Paginated[T](results: Seq[T], page: Page, total: Long)
+object Paginated {
+  def apply[T](results: T*): Paginated[T] = Paginated[T](results, Page(), results.size)
+  def empty[T](page: Page) = Paginated(Seq.empty[T], page, 0)
+
+  implicit def writes[T](implicit fmt: Writes[T]): Writes[Paginated[T]] = new Writes[Paginated[T]] {
```
ed0906: Generics means doing custom `Writes`....

app/uk/gov/hmrc/exports/models/Page.scala
bambuchaAdm: I don't think we have to take implicit argument here - this `QueryStringBindable[Int]` could be field in `QueryStringBindable[Page]` with usage of `implicitly` language function

app/uk/gov/hmrc/exports/models/Page.scala
```diff
@@ -0,0 +1,45 @@
...
+    new QueryStringBindable[Page] {
+      private def queryParamIndex(key: String) = key + "-index"
+      private def queryParamSize(key: String) = key + "-size"
+
+      override def bind(key: String, params: Map[String, Seq[String]]): Option[Either[String, Page]] = {
+        val index: Option[Int] = params.get(queryParamIndex(key)).flatMap(_.headOption).map(_.toInt)
+        val size: Option[Int] = params.get(queryParamSize(key)).flatMap(_.headOption).map(_.toInt)
+        (index, size) match {
```
bambuchaAdm: maybe just getOrElse on options instead of matching ?

ed0906: I kind of wanted the default values to live inside `Page` so we only have them in one place

ed0906: I guess I could make them fields that I can access from here

ed0906: Fixed

app/uk/gov/hmrc/exports/models/Page.scala
ed0906: Nice

ed0906: Fixed

PR #76, opened 2019-08-16
CEDS Improve Logging

ed0906:


app/uk/gov/hmrc/exports/controllers/DeclarationController.scala
```diff
@@ -87,4 +97,9 @@ class DeclarationController @Inject()(
     }
   }
 
+  private def logPayload[T](prefix: String, t: T)(implicit wts: Writes[T]): T = {
```
PatrykRudnicki: The same like on frontend. Ask for change `t` to `payload`

PR #77, opened 2019-08-16
CEDS-1484 Add Submission & Notifications endpoint

ed0906:


conf/app.routes
```diff
@@ -1,10 +1,12 @@
...
-DELETE      /v2/declarations/:id                              uk.gov.hmrc.exports.controllers.DeclarationController.deleteByID(id: String)
+POST        /v2/declarations                                 uk.gov.hmrc.exports.controllers.DeclarationController.create()
+GET         /v2/declarations                                 uk.gov.hmrc.exports.controllers.DeclarationController.findAll(status: Option[String] ?= None, page: uk.gov.hmrc.exports.models.Page ?= uk.gov.hmrc.exports.models.Page())
+PUT         /v2/declarations/:id                             uk.gov.hmrc.exports.controllers.DeclarationController.update(id: String)
+GET         /v2/declarations/:id                             uk.gov.hmrc.exports.controllers.DeclarationController.findByID(id: String)
+DELETE      /v2/declarations/:id                             uk.gov.hmrc.exports.controllers.DeclarationController.deleteByID(id: String)
+GET         /v2/declarations/:id/submission                  uk.gov.hmrc.exports.controllers.SubmissionController.findByID(id: String)
+GET         /v2/declarations/:id/submission/notifications    uk.gov.hmrc.exports.controllers.NotificationController.findByID(id: String)
```
ed0906: These should not allow us to get rid of the dependency on Conversation ID

app/uk/gov/hmrc/exports/controllers/JSONResponses.scala
```diff
@@ -0,0 +1,30 @@
...
+ * limitations under the License.
+ */
+
+package uk.gov.hmrc.exports.controllers
+
+import play.api.http.{ContentTypeOf, ContentTypes, Writeable}
+import play.api.libs.json.Writes
+import play.api.mvc.Codec
+
+trait JSONResponses {
```
PatrykRudnicki: I think putting this in trait and extending every Controller with this it's not a good idea, but it's just my opinion.

app/uk/gov/hmrc/exports/controllers/NotificationController.scala
```diff
@@ -42,17 +41,27 @@ class NotificationController @Inject()(
...
+          .map(notifications => Ok(notifications))
+      case _ => Future.successful(NotFound)
+    }
+  }
+
   def getSubmissionNotifications(mrn: String): Action[AnyContent] =
     authorisedAction(bodyParsers.default) { implicit request =>
       notificationsService
         .getNotificationsForSubmission(mrn)
-        .map(notifications => Ok(Json.toJson(notifications)))
```
PatrykRudnicki: And I prefer previous explicit `Json.toJson` I'm not sure that you're solving some problem or it's just improvement to write less code

ed0906: Personally if I don't have to write `Json.toJson` each time its better

PatrykRudnicki: Yeah, but for new Controller you need to remember about this Trait, I'm not sure about it, let's wait for other people :) I don't want to block PR :)

ed0906: New controllers will extend the RESTController trait which pulls it in for free. I didnt want to change this one as it's likely to be removed once we stop using conversation-id everywhere

mattclark-zerogravit: I agree with Patryk, its not obvious you need to include this trait. Also the Json.toJson whilst more verbose shows there is a conversion actually happening. We are invoking the Json writer here. Removing it hides that and it all just happens magically which reads nicer but could argue is more confusing.

ed0906: For me less boiler plate is better, Play is the only framework I've seen that makes you manually convert all payloads to & from strings, others e.g. SpringBoot, DropWizard presume you want JSON and do it for you.
 
Regarding creating new controllers, it's easy to miss a lot of things so we'd generally copy and paste an existing one anyway. Any new Controllers would extend `RESTController` which gives everything for free

SteveSugden: If the DeclarationController (which extends RESTController) is the accepted "new" way of creating json responses then I see no reason why the Notification and Submission controllers shouldn't (temporarily?) do the same thing.

PR #80, opened 2019-08-19
CEDS-1507 - Error with date at exports frontend

SteveSugden:


app/uk/gov/hmrc/exports/services/mapping/governmentagencygoodsitem/AdditionalDocumentsBuilder.scala
```diff
@@ -170,7 +170,7 @@ object AdditionalDocumentsBuilder {
             DateTimeElement(
               DateTimeString(
                 formatCode = dateTimeCode,
-                value = LocalDate.parse(date.toString).format(DateTimeFormatter.ofPattern("yyyyMMdd"))
+                value = date.toLocalDate.format(DateTimeFormatter.ofPattern("yyyyMMdd"))
```
bambuchaAdm: Can we put his formatter in companion object ?

SteveSugden: Could have.  But that wasn't the problem and I've merged now

PR #83, opened 2019-08-20
CEDS-1487 - ability to specify sort order of find declarations

SteveSugden:
Adds the ability to specify a sort order and direction when calling the GET declarations end-point.
Required for CEDS-1487 which needs declarations in the order they were last modified (most recent first)


SteveSugden:
test this please

SteveSugden:
test this please

app/uk/gov/hmrc/exports/controllers/DeclarationController.scala
```diff
@@ -72,13 +72,20 @@ class DeclarationController @Inject()(
       }
     }
 
-  def findAll(status: Option[String], pagination: Page): Action[AnyContent] =
+  def findAll(
+    status: Option[String],
+    pagination: Page,
+    sortBy: Option[String],
```
ed0906: Any chance we can wrap these in a `Sort` class a bit like `Page`?

ed0906: It would mean on other controllers that might use this we can reuse all the parsing/validation code

ed0906: I see you've got a `DeclarationSort` class lower down. You could use `Binder` e.g in `Page` to have that as a method parameter

app/uk/gov/hmrc/exports/repositories/SubmissionRepository.scala
```diff
@@ -43,13 +43,17 @@ class SubmissionRepository @Inject()(implicit mc: ReactiveMongoComponent, ec: Ex
   override def indexes: Seq[Index] = Seq(
     Index(Seq("actions.conversationId" -> IndexType.Ascending), unique = true, name = Some("conversationIdIdx")),
     Index(Seq("eori" -> IndexType.Ascending), name = Some("eoriIdx")),
-    Index(Seq("eori" -> IndexType.Ascending, "action.requestTimestamp" -> IndexType.Descending), name = Some("actionOrderedEori"))
+    Index(
+      Seq("eori" -> IndexType.Ascending, "action.requestTimestamp" -> IndexType.Descending),
```
ed0906: 👍

app/uk/gov/hmrc/exports/repositories/SubmissionRepository.scala
```diff
@@ -43,13 +43,17 @@ class SubmissionRepository @Inject()(implicit mc: ReactiveMongoComponent, ec: Ex
   override def indexes: Seq[Index] = Seq(
     Index(Seq("actions.conversationId" -> IndexType.Ascending), unique = true, name = Some("conversationIdIdx")),
     Index(Seq("eori" -> IndexType.Ascending), name = Some("eoriIdx")),
-    Index(Seq("eori" -> IndexType.Ascending, "action.requestTimestamp" -> IndexType.Descending), name = Some("actionOrderedEori"))
+    Index(
```
ed0906: We probably need another index for your sort field

conf/app.routes
```diff
@@ -1,7 +1,7 @@
 # microservice specific routes
 
 POST        /v2/declarations                                 uk.gov.hmrc.exports.controllers.DeclarationController.create()
-GET         /v2/declarations                                 uk.gov.hmrc.exports.controllers.DeclarationController.findAll(status: Option[String] ?= None, page: uk.gov.hmrc.exports.models.Page ?= uk.gov.hmrc.exports.models.Page())
+GET         /v2/declarations                                 uk.gov.hmrc.exports.controllers.DeclarationController.findAll(status: Option[String] ?= None, page: uk.gov.hmrc.exports.models.Page ?= uk.gov.hmrc.exports.models.Page(), sortBy: Option[String] ?= None, sortDirection: Option[String] ?= None)
```
ed0906: Would be nicer if we could bind the `DeclarationSort` class here rather than two params

SteveSugden: I noticed how that was done with the 'Page'.  Good suggestion.

test/util/testdata/ExportsDeclarationBuilder.scala
```diff
@@ -272,6 +272,9 @@ trait ExportsDeclarationBuilder {
       )
     )
 
+  def withUpdateTime(year: Int, month: Int, dayOfMonth: Int): ExportsDeclarationModifier =
```
ed0906: `withUpdateDate`? given it takes a year/month/day

app/uk/gov/hmrc/exports/models/DeclarationSort.scala
```diff
@@ -0,0 +1,24 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package uk.gov.hmrc.exports.models
+
+case class DeclarationSort(by: String = "createdDateTime", direction: Int = DeclarationSort.ASC)
```
ed0906: It'd be nice if we could keep the `by` and `direction` fields in an enum, that way we can protect against sorting on unexpected fields.

e.g. if I try to sort on "id" or something, the binder might throw a 400 or just use the default sort order

SteveSugden: Is that because sorting on any random field is in-efficient without an index on that field?

app/uk/gov/hmrc/exports/controllers/DeclarationController.scala
```diff
@@ -72,13 +72,20 @@ class DeclarationController @Inject()(
...
+    sortDirection: Option[String]
+  ): Action[AnyContent] =
     authenticator.authorisedAction(parse.default) { implicit request =>
       val search = DeclarationSearch(
         eori = request.eori.value,
         status = status.map(v => Try(DeclarationStatus.withName(v))).filter(_.isSuccess).map(_.get)
       )
-      declarationService.find(search, pagination).map(results => Ok(results))
+      def direction: Option[String] => Int = option => if (option.exists(_.equalsIgnoreCase("asc"))) 1 else -1
+      val sort = sortBy.map(field => DeclarationSort(field, direction(sortDirection))).getOrElse(DeclarationSort())
```
bambuchaAdm: Can we move `direction` function as apply in `DeclarationSort

app/uk/gov/hmrc/exports/models/DeclarationSort.scala
bambuchaAdm: Implementation of `QueryStringBindable` type class would be nice

ed0906: > Is that because sorting on any random field is in-efficient without an index on that field?

Yehh there is that, but also for a validation point of view

app/uk/gov/hmrc/exports/controllers/DeclarationController.scala
ed0906: I think he's pulling it into a binder

app/uk/gov/hmrc/exports/controllers/DeclarationController.scala
SteveSugden: Done

app/uk/gov/hmrc/exports/repositories/SubmissionRepository.scala
SteveSugden: Done - can you check @ed0906

conf/app.routes
SteveSugden: Done

app/uk/gov/hmrc/exports/repositories/SubmissionRepository.scala
ed0906: Looks good, presume thats the only field we need to sort on for the moment right?

I think `status` needs one too actually

SteveSugden: Add an index for status or the ability to sort-by?   That can be added when we switch to using the new findDeclaration service I guess

SteveSugden:
test this please

PR #86, opened 2019-08-21
CEDS-1528 Fix Component test after /v2/declarations behaviour change

PauloPortugal:


test/component/uk/gov/hmrc/exports/ExportsSubmissionReceivedSpec.scala
```diff
@@ -17,240 +17,289 @@
...
+      withDeclarationRepositorySuccess()
+
+      startSubmissionService(ACCEPTED)
+      val request: FakeRequest[AnyContentAsJson] = ValidSubmissionRequest
+
+      Given("user is authorised")
+      authServiceAuthorizesWithEoriAndNoRetrievals()
+
+      When("a POST request with data is sent to the API")
+      val result = route(app = app, request).value
```
PatrykRudnicki: In my opinion there is no value to use `app = app` instead of `app`

test/component/uk/gov/hmrc/exports/ExportsSubmissionReceivedSpec.scala
```diff
@@ -17,240 +17,289 @@
...
+
+      Then(s"a response with a $CREATED status is received")
```
PatrykRudnicki: Is there any reason to use `$CREATED` instead just a string? Without this you don't need to use `s`.

test/component/uk/gov/hmrc/exports/ExportsSubmissionReceivedSpec.scala
PauloPortugal: This has been addressed.

test/component/uk/gov/hmrc/exports/ExportsSubmissionReceivedSpec.scala
PauloPortugal: This has been addressed.

PR #87, opened 2019-08-27
CEDS-1516 - purge stale draft declarations

SteveSugden:


app/uk/gov/hmrc/exports/scheduler/PurgeDraftDeclarationsJob.scala
```diff
@@ -0,0 +1,54 @@
...
+
+  override val name: String = "PurgeDraftDeclarations"
+  override def interval: FiniteDuration = jobConfig.interval
+  override def firstRunTime: LocalTime = jobConfig.elapseTime
+
+  override def execute(): Future[Unit] = {
+
+    val expiryDate = Instant.now(clock).minusSeconds(expireDuration.toSeconds)
+    for {
+      result <- declarationRepository.deleteExpiredDraft(expiryDate)
```
bambuchaAdm: why not just
```
declarationRepository.deleteExpiredDraft(expiryDate)
  .andThen{ case _ => Logger.info(s"${name}Job: Purged ${result.n} items updated before $expiryDate"))
```

SteveSugden: That would be simpler, yes.

SteveSugden: Actually, to preserve the method signature and handle the failure case, it's turned into the following..
```
    declarationRepository
      .deleteExpiredDraft(expiryDate)
      .andThen {
        case Success(result) => Logger.info(s"${name}Job: Purged ${result.n} drafts updated before $expiryDate")
        case Failure(t)      => Logger.error(s"Error occurred purging drafts updated before $expiryDate  [${t.getMessage}]")
      }
      .map(_ => Unit)
```

test/unit/uk/gov/hmrc/exports/scheduler/SchedulerDateUtilSpec.scala
```diff
@@ -0,0 +1,74 @@
...
+    "Calculate the next run date given a run time of now" in {
+      util.nextRun(time("12:00"), 3.seconds) shouldBe instant("2019-01-01T12:00:00")
+    }
+
+    "Calculate the next run date given a run time in the future" in {
+      util.nextRun(time("12:00:09"), 3.seconds) shouldBe instant("2019-01-01T12:00:00")
+    }
+
+    "Calculate the next run date given a run time in the future with Offset" in {
+      util.nextRun(time("12:00:01"), 3.seconds) shouldBe instant("2019-01-01T12:00:01")
```
bambuchaAdm: This line is very similar to `Calculate the next run date given a run time in the future" in` -  why outcome is differ ?

SteveSugden: In the first test (`Calculate the next run date given a run time in the future`) we are stating that we want to run at `12:00:09` with an interval of 3 seconds.   So the current time (`12:00:00`) is a valid next-run-time as the process will run at `12:00:09` after 3x3secs interval.

In the second test (`Calculate the next run date given a run time in the future with Offset`) - first example, we want the scheduler to run at `12:00:01` with an interval of 3 secs.  As the current time is `12:00:00` the next valid run time is `12:00:01`.

Hope that makes sense - had to get Ed to explain it to me.

app/uk/gov/hmrc/exports/scheduler/PurgeDraftDeclarationsJob.scala
SteveSugden: Does the way I wrote it first time mean that I don't have to worry about/handle the failure case?  I can leave it to the caller to log the error.

app/uk/gov/hmrc/exports/config/AppConfig.scala
```diff
@@ -52,4 +58,14 @@ class AppConfig @Inject()(
 
   lazy val developerHubClientId: String =
     servicesConfig.getString("microservice.services.customs-declarations.client-id")
+
+  lazy val draftTimeToLive: FiniteDuration =
+    servicesConfig.getDuration("draft.timeToLive").asInstanceOf[FiniteDuration]
+
+  lazy val purgeDraftDeclarations: JobConfig = JobConfig(
```
ed0906: Think this needs `lifetime` or `interval` or something adding to its name

app/uk/gov/hmrc/exports/repositories/DeclarationRepository.scala
```diff
@@ -81,4 +84,10 @@ class DeclarationRepository @Inject()(mc: ReactiveMongoComponent, appConfig: App
       .remove("id" -> declaration.id, "eori" -> declaration.eori)
       .map(_ => Unit)
 
+  def deleteExpiredDraft(expiryDate: Instant): Future[WriteResult] = {
```
ed0906: 👍

ed0906: Please can we not expose `WriteResult` to the service layer as it is a database layer concern

It should have a count field on it which we can return, otherwise I'd suggest returning `Future[Unit]`

app/uk/gov/hmrc/exports/config/AppConfig.scala
SteveSugden: The JobConfig val name?  How about `purgeDraftDeclarationsJob`?  it contains both the run-time and interval.

ed0906: Ignore me, just noticed it was a wrapper around two config variables

app/uk/gov/hmrc/exports/repositories/DeclarationRepository.scala
SteveSugden: Good point.  It's just the count field I'm interested in (for logging)

SteveSugden: ...alternatively, the repo class could log the outcome (i.e. how many records deleted)

ed0906: I'd just expose the count with `Future[Int]`

app/uk/gov/hmrc/exports/scheduler/Scheduler.scala
```diff
@@ -0,0 +1,67 @@
...
+@Singleton
+class Scheduler @Inject()(
+  actorSystem: ActorSystem,
+  appConfig: AppConfig,
+  schedulerDateUtil: SchedulerDateUtil,
+  scheduledJobs: ScheduledJobs
+) {
+
+  scheduledJobs.jobs.foreach { job =>
+    Logger.info(
```
PatrykRudnicki: I think it'll be good to create logger for Scheduler class and use it, not this from play, because when you create your own logger the log will contain whole information what the log is it, so in this case you'll know that it's Scheduler log.

`private val logger = Logger(classOf[Scheduler])`

app/uk/gov/hmrc/exports/repositories/DeclarationRepository.scala
SteveSugden: That's what I went with

app/uk/gov/hmrc/exports/scheduler/Scheduler.scala
SteveSugden: Good spot.  Thanks

ed0906: I think we need to agree on our approach to this team wide as I've noticed a few different approaches to logging, could you stick a note on the tech debt board please @PatrykRudnicki  :)

ed0906:
test this please

app/uk/gov/hmrc/exports/scheduler/Scheduler.scala
PatrykRudnicki: Yes, we had this metting before, but we can do another one, because team changed :)

app/uk/gov/hmrc/exports/scheduler/PurgeDraftDeclarationsJob.scala
```diff
@@ -0,0 +1,54 @@
...
+  override val name: String = "PurgeDraftDeclarations"
+  override def interval: FiniteDuration = jobConfig.interval
+  override def firstRunTime: LocalTime = jobConfig.elapseTime
+
+  override def execute(): Future[Unit] = {
+
+    val expiryDate = Instant.now(clock).minusSeconds(expireDuration.toSeconds)
+    for {
+      count <- declarationRepository.deleteExpiredDraft(expiryDate)
+      _ = Logger.info(s"${name}Job: Purged $count items updated before $expiryDate")
```
PatrykRudnicki: The same here will logger, but we can wait for discussion about this

test/unit/uk/gov/hmrc/exports/scheduler/PurgeDraftDeclarationsJobSpec.scala
```diff
@@ -0,0 +1,84 @@
...
+
+  private val zone = ZoneOffset.UTC
+  private val clock: Clock = Clock.fixed(Instant.now(), zone)
+  private val appConfig = mock[AppConfig]
+  private val declarationRepository = mock[DeclarationRepository]
+
+  override def afterEach(): Unit = {
+    super.afterEach()
+    reset(appConfig, declarationRepository)
+    given(appConfig.clock) willReturn clock
```
PatrykRudnicki: We're using `when` `thenReturn` for unit specs, but it's not an issue

test/unit/uk/gov/hmrc/exports/scheduler/SchedulerSpec.scala
```diff
@@ -0,0 +1,132 @@
...
+  }
+
+  private implicit def instant2Time: Instant => LocalTime = _.atZone(zone).toLocalTime
+
+  override protected def beforeEach(): Unit = {
+    super.beforeEach()
+    given(config.clock) willReturn clock
+    given(actorSystem.scheduler) willReturn internalScheduler
+    given(internalScheduler.schedule(any[FiniteDuration], any[FiniteDuration], any[Runnable])(any[ExecutionContext])) will runTheJobImmediately
+    given((job.execute())) willReturn Future.successful(())
```
PatrykRudnicki: Double unnecessary parenthesis `given((job.execute()))` -> `given(job.execute())`

test/unit/uk/gov/hmrc/exports/scheduler/PurgeDraftDeclarationsJobSpec.scala
ed0906: I don't think we should be too picky about this kind of thing, both are readable and theres no real benefit of being consistent

PR #89, opened 2019-08-29
CEDS-1552 Consistent way of using loggers

PatrykRudnicki:


app/uk/gov/hmrc/exports/controllers/RESTController.scala
```diff
@@ -26,14 +26,16 @@ import scala.concurrent.ExecutionContext
 abstract class RESTController(override val controllerComponents: ControllerComponents)
     extends BackendController(controllerComponents) with JSONResponses {
 
+  private val logger = Logger(this.getClass)
```
bambuchaAdm: why private here when you use dynamic 'getClass' ?

PatrykRudnicki: It's based on what behaviour you want to achieve with this.

For this I can change it to specific logger for the `RESTController`.

`Logger("RESTController")`

PR #91, opened 2019-09-04
BUG - excessive d.p. when converting non-integer StatisticalValueAmount

SteveSugden:


app/uk/gov/hmrc/exports/services/WcoSubmissionService.scala
```diff
@@ -41,6 +44,8 @@ class WcoSubmissionService @Inject()(
       .getOrElse(throw new IllegalArgumentException("An DUCR is required"))
     val payload = wcoMapperService.toXml(metaData)
 
+   logger.debug(s"Sending $payload for $declaration")
```
SteveSugden: Is this allowed/advisable?   Beth suggested it would be useful to see what xml we are sending.

app/uk/gov/hmrc/exports/services/mapping/governmentagencygoodsitem/StatisticalValueAmountBuilder.scala
```diff
@@ -31,18 +31,14 @@ class StatisticalValueAmountBuilder @Inject()() extends ModifyingBuilder[ExportI
   def buildThenAdd(exportItem: ExportItem, wcoGovernmentAgencyGoodsItem: WCOGovernmentAgencyGoodsItem): Unit =
     exportItem.itemType.foreach { itemType =>
       wcoGovernmentAgencyGoodsItem.setStatisticalValueAmount(
-        createWCODecStatisticalValueAmount(Some(BigDecimal(itemType.statisticalValue)), Some(defaultCurrencyCode))
+        createWCODecStatisticalValueAmount(itemType.statisticalValue, defaultCurrencyCode)
       )
     }
```
SteveSugden: Got rig of the useless string -> optional scala BD -> non-optional java BD conversions here

app/uk/gov/hmrc/exports/services/WcoSubmissionService.scala
PauloPortugal: I would not do that, because this will contain PII and then we will require to clean up the logs when we need it in production.
You can create a `printNonPII` method that prints what you all the payload without PII "EORI, ADDRESSES, NAMES, etc"

PauloPortugal: First we need to understand what is PII, which I believe a EORI is one, but Ian is asking that

SteveSugden: I wondered about that.  Beth said we dump the dec json at the frontend.   I'll get rid of logging then.

PauloPortugal: we will have to stop dumping the payload... unless we filter it

SteveSugden: I've removed the logging.   If its needed we can create a ticket.   Was just a (poor) idea of mine

bambuchaAdm: On level of debug should be fine. This is exactly like logging guide suggest

PR #102, opened 2019-09-24
Index fixes

bambuchaAdm:
Empty actions array in submission were indexed as null for `action.id` index. 
Moving `action.id` index as partial index only for document that have at least one action.

### DIFF
--- a/app/uk/gov/hmrc/exports/repositories/SubmissionRepository.scala
+++ b/app/uk/gov/hmrc/exports/repositories/SubmissionRepository.scala
@@ -22,7 +22,7 @@ import play.modules.reactivemongo.ReactiveMongoComponent
 import reactivemongo.api.Cursor.FailOnError
 import reactivemongo.api.ReadPreference
 import reactivemongo.api.indexes.{Index, IndexType}
-import reactivemongo.bson.BSONObjectID
+import reactivemongo.bson.{BSONBoolean, BSONDocument, BSONObjectID}
 import reactivemongo.play.json.commands.JSONFindAndModifyCommand.FindAndModifyResult
 import uk.gov.hmrc.exports.models.Eori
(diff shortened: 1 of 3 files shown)


app/uk/gov/hmrc/exports/repositories/SubmissionRepository.scala
```diff
@@ -42,7 +42,12 @@ class SubmissionRepository @Inject()(implicit mc: ReactiveMongoComponent, ec: Ex
     ) {
 
   override def indexes: Seq[Index] = Seq(
-    Index(Seq("actions.id" -> IndexType.Ascending), unique = true, name = Some("actionIdIdx")),
+    Index(
+      Seq("actions.id" -> IndexType.Ascending),
+      unique = true,
+      name = Some("actionIdIdx"),
+      partialFilter = Some(BSONDocument.apply(Seq("action.id" -> BSONDocument.apply("$exist" -> BSONBoolean(true)))))
```
ed0906: `BSONDocument(...)` not `BSONDocument.apply(...)`

PatrykRudnicki:
Scala formatter

test/integration/uk/gov/hmrc/exports/repositories/SubmissionRepositorySpec.scala
```diff
@@ -87,6 +87,12 @@ class SubmissionRepositorySpec
         submissionsInDB.head must equal(submission)
       }
     }
+
+    "allow save two submissions with empty actions" in {
```
ed0906: Please can you prove this index works with a negative test as well
e.g. add two actions with the same ID and expect an exception

bambuchaAdm: Covered at line 68 in `trying to save Submission with the same conversationId twice throw DatabaseException`

PR #105, opened 2019-09-30
CEDS-1749 Add indexes on declarations repository

bambuchaAdm:


test/integration/uk/gov/hmrc/exports/repositories/DeclarationRepositoryTest.scala
```diff
@@ -56,7 +56,7 @@ class DeclarationRepositoryTest
       .toInt
 
   private def givenADeclarationExists(declarations: ExportsDeclaration*): Unit =
-    repository.collection.insert(false).many(declarations).futureValue
+    repository.collection.insert(ordered = true).many(declarations).futureValue
```
ed0906: Why is this necessary?

test/util/testdata/ExportsDeclarationBuilder.scala
```diff
@@ -53,7 +54,7 @@ trait ExportsDeclarationBuilder {
   )
 
   def aDeclaration(modifiers: ExportsDeclarationModifier*): ExportsDeclaration =
-    modifiers.foldLeft(modelWithDefaults)((current, modifier) => modifier(current))
+    modifiers.foldLeft(modelWithDefaults.copy(id = uuid))((current, modifier) => modifier(current))
```
ed0906: if `modelWithDefaults` with a def then you wouldn't need this

test/integration/uk/gov/hmrc/exports/repositories/DeclarationRepositoryTest.scala
bambuchaAdm: yes - test depends on strict order of fetched documents - this change is only in tests

test/util/testdata/ExportsDeclarationBuilder.scala
ed0906: I'm pretty sure it used to be a `def` for this reason

PR #111, opened 2019-10-09
CEDS-1695 WCO to Exports Pointer Mapping

ed0906:
This maps notifications with error blocks e.g.
```
<_2_1:Error>
        <_2_1:ValidationCode>CDS10020</_2_1:ValidationCode>
        <_2_1:Pointer>
          <_2_1:DocumentSectionCode>42A</_2_1:DocumentSectionCode>
          <_2_1:TagID>D026</_2_1:TagID> FunctionalReferenceID
        </_2_1:Pointer>
</_2_1:Error>
```
To our notification model with a "pointer" field which corresponds to what that field is in our model

```
"actionId": "...",
// .... other Notification fields...
"errors" : [
        {
            "validationCode" : "CDS10020",
            "pointer" : "declaration.consignmentReferences.lrn"
        }
    ]
```

This also works when the xml `Pointer` refers to a Sequence
In this case our Notification model will look like:
```
"actionId": "...",
// .... other Notification fields...
"errors" : [
        {
            "validationCode" : "CDS10020",
            "pointer" : "declaration.consignmentReferences.0.lrn"
        }
    ]
```
Where the 0 would be the index in the `ConsignmentReferences` sequence that the error applies to.

### How this can be used in the front end to link to a field
In the front end we can then use the `"declaration.consignmentReferences.0.lrn"` by doing something along the lines of:
- Swap out the numerical sections with a `$` or other symbol e.g. `"declaration.consignmentReferences.$.lrn"` 
- Have a Messages file entry for that which is the Field Name (Phase 1)


app/uk/gov/hmrc/exports/services/WCOPointerMappingService.scala
```diff
@@ -0,0 +1,12 @@
+package uk.gov.hmrc.exports.services
+
+import javax.inject.Inject
+import uk.gov.hmrc.exports.models.Pointer
+
+class WCOPointerMappingService @Inject()(){
```
ed0906: So this is the entry point to the Pointer Mapping

If you can construct a WCO `Pointer` from the Notification XML, pass it to this method and it will convert it to the exports equivalent.

I've added a method that accepts a `Seq[Pointer]` if it's useful (depends if you want to batch up the mapping or not), we can delete it if not.

app/uk/gov/hmrc/exports/models/Pointer.scala
```diff
@@ -0,0 +1,27 @@
...
+  // e.g. ABC.DEF.GHI (if the pointer doesnt contain a sequence)
+  def pattern: String = sections.map(_.pattern).mkString(".")
+
+  // Converts a pointer into its value form
+  // e.g. ABC.DEF.1.GHI  (if the pointer contains a sequence with index 1)
+  // e.g. ABC.DEF.GHI (if the pointer doesnt contain a sequence)
+  def value: String = sections.map(_.value).mkString(".")
+}
+
+case class PointerSection(value: String, `type`: WCOPointerSectionType) {
```
ed0906: This will represent one Pointer section in the XML e.g.
```
<Pointer>
    <DocumentSectionCode>42A</DocumentSectionCode>
</Pointer>
```
is equivalent to `PointerSection("42A", PointerSectionType.FIELD)`

ed0906: For sequences the XML seems to look like
```
<Pointer>
    <SequenceNumeric>1</SequenceNumeric>
    <DocumentSectionCode>68A</DocumentSectionCode>
</Pointer>
```
This would be two `PointerSection`s
`PointerSection("1", PointerSectionType.SEQUENCE)`
`PointerSection("68A", PointerSectionType.FIELD)`

ed0906: For XML nodes with `tagID`s these would be separate `PointerSection`s too
```
<Pointer>
    <DocumentSectionCode>70A</DocumentSectionCode>
    <TagID>166</TagID>
</Pointer>
```
If I understand correctly the `TagID` is just the final section of the error pointer, lets just treat it like any other section
`PointerSection("70A", PointerSectionType.FIELD)`
`PointerSection("166", PointerSectionType.FIELD)`

app/uk/gov/hmrc/exports/models/Pointer.scala
```diff
@@ -0,0 +1,27 @@
+package uk.gov.hmrc.exports.models
+
+import uk.gov.hmrc.exports.models.PointerSectionType.WCOPointerSectionType
+
+case class Pointer(sections: Seq[PointerSection]) {
```
ed0906: This will represent the entire pointer in the XML (multiple `Pointer` xml nodes).  The existing model models it as `Seq[ErrorPointer]`.  
I think the existing model will be awkward to work with in terms of the mapping as we're going to need a tonne of helper methods and the knowledge of wether it's a sequence element or not

app/uk/gov/hmrc/exports/services/NotificationService.scala
```diff
@@ -87,4 +92,84 @@ class NotificationService @Inject()(
...
+        */
+      val tagId: Option[PointerSection] =
+        if ((singlePointerXml \ "TagID").nonEmpty)
+          Some(PointerSection((singlePointerXml \ "TagID").text, PointerSectionType.FIELD))
+        else None
+
+      List(documentSectionCode, sequenceNumeric, tagId).flatten
+    }.toList
+
+    Pointer(pointerSections)
```
ed0906: Basically @PatrykRudnicki this is where you would call the `WCOPointerMappingService` to convert that `Pointer` to an Exports one

PatrykRudnicki: Yes, I changed it

ed0906:
test this please

app/uk/gov/hmrc/exports/models/PointerMapping.scala
```diff
@@ -0,0 +1,55 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package uk.gov.hmrc.exports.models
+
+case class PointerMapping(wcoPattern: PointerPattern, exportsPattern: PointerPattern) {
+
+  def applyToWCOPointer(pointer: Pointer): Pointer = {
+    if (!wcoPattern.matches(pointer.pattern))
```
PauloPortugal: How are we thinking of obtaining all the wco patterns?

app/uk/gov/hmrc/exports/services/WCOPointerMappingService.scala
```diff
@@ -0,0 +1,42 @@
...
+
+package uk.gov.hmrc.exports.services
+
+import javax.inject.Inject
+import uk.gov.hmrc.exports.models.{Pointer, PointerMapping, PointerPattern}
+import uk.gov.hmrc.exports.util.FileReader
+
+class WCOPointerMappingService @Inject()(fileReader: FileReader) {
+
+  private lazy val mappings: Set[PointerMapping] = {
```
PauloPortugal: Very good! 
I would have suggested a Json structure as it helps understand what are we mapping more clearly and it also allows for better flexibility and allows for better deserialisation support, i.e, imagine if we want to provide error messages or other information, probably better than having to splic by comma. Probably a YAGNI at this point.

app/uk/gov/hmrc/exports/services/WCOPointerMappingService.scala
```diff
@@ -0,0 +1,42 @@
...
+      .readLines("code-lists/pointer-mapping.csv")
+      .map(_.split(","))
+      .map {
+        case Array(wcoPattern, exportsPattern) =>
+          PointerMapping(PointerPattern(wcoPattern.trim), PointerPattern(exportsPattern.trim))
+      }
+      .toSet
+  }
+
+  def mapWCOPointerToExportsPointer(pointer: Pointer): Option[Pointer] =
```
PauloPortugal: :+1:

app/uk/gov/hmrc/exports/models/Pointer.scala
```diff
@@ -0,0 +1,103 @@
...
+}
+
+case class Pointer(sections: Seq[PointerSection]) {
+  //  Converts a pointer into it's pattern form
+  // e.g. ABC.DEF.*.GHI (if the pointer contains a sequence index)
+  // e.g. ABC.DEF.GHI (if the pointer doesnt contain a sequence)
+  lazy val pattern: PointerPattern = PointerPattern(sections.map(_.pattern).mkString("."))
+
+  // Converts a pointer into its value form
+  // e.g. ABC.DEF.1.GHI  (if the pointer contains a sequence with index 1)
```
PauloPortugal: Is it possible to have  `ABC.DEF.2.GHI` or  `ABC.DEF.4.GHI`?
In what way would the field differ from a `ABC.DEF.GHI`?
Would it just be to point to the user which element is causing an issue?

app/uk/gov/hmrc/exports/models/PointerMapping.scala
ed0906: Steve G has a mapping file, perhaps we can use the work you did in the wco-dec eventually to at least generate the left hand side of the mapping (`42A.26B.123.ABC` bit)
That way we'd know if we need to be handling it based on the schema

app/uk/gov/hmrc/exports/services/WCOPointerMappingService.scala
ed0906: Good suggestion, it'd be easy to refactor in as we could just write a Json Formatter for `PointerMapping`

For the purpose of this story though I think its okay for now

SteveSugden: Not withstanding the possibility of moving to Json - did Patryk not introduce some CSV helper package recently?

PR #112, opened 2019-10-11
CEDS-1695 Fix for Pointer Fields that have numeric codes

ed0906:
We used to have a custom json formatter that assumed that because a pointer section was numerical, then it represented a sequence index. This is not the case, some pointer field references are purely numerical and we need to handle that.

Sequence sections are now rendered in JSON as `#1` (for index 1)
e.g.
Pointer Before: `abc.1.def`
Pointer After: `abc.#1.def`

It means we can convert to and from the string/json version and retain the "section type"

### DIFF
--- a/app/uk/gov/hmrc/exports/models/Pointer.scala
+++ b/app/uk/gov/hmrc/exports/models/Pointer.scala
@@ -25,8 +25,7 @@ import scala.util.{Success, Try}
 object PointerSectionType extends Enumeration {
   type PointerSectionType = Value
   val FIELD, SEQUENCE = Value
-  implicit val format: Format[models.PointerSectionType.Value] =
-    Format(Reads.enumNameReads(PointerSectionType), Writes.enumNameWrites)
+  implicit val format: Format[models.PointerSectionType.Value] = Format(Reads.enumNameReads(PointerSectionType), Writes.enumNameWrites)
 }
 
 case class PointerSection(value: String, `type`: PointerSectionType) {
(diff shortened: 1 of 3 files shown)


app/uk/gov/hmrc/exports/models/Pointer.scala
```diff
@@ -25,19 +25,30 @@ import scala.util.{Success, Try}
...
+  }
 }
 
 object PointerSection {
-  implicit val format = Json.format[PointerSection]
+  private val SEQUENCE_REGEX = "^#(\\d*)$".r
+  implicit val format: Format[PointerSection] =
+    Format[PointerSection](Reads(js => js.validate[String].map(PointerSection(_))), Writes(section => JsString(section.toString)))
+
+  def apply(value: String): PointerSection = SEQUENCE_REGEX.findFirstMatchIn(value).map(_.group(1)) match {
```
PatrykRudnicki: What this 1 means? 
`The matched string in group i, or null if nothing was matched.`

This is from documentation, do we need this `group(1)`?

What do you thinki about using `matched`?
https://www.scala-lang.org/api/2.12.1/scala/util/matching/Regex$$Match.html#matched:String

ed0906: Tbh a Regex was probably a bit overkill for this... All It is doing is checking if the string starts with a `#` and then it extracts the number after it if it does

ed0906: Could easily be `.filter(_.startsWith("#")).map(_.replaceAll("#", ""))` I guess

ed0906: Or something along those lines

PR #114, opened 2019-10-21
CEDS-1718 Rename Choice to DeclarationType

ed0906:
Currently the `ExportsDeclaration` model has a choice field which is a `String`.
Given the many extra dec types we have to add over the next few months we should make this easy to expand. 

`choice` seems the wrong name for this & `String` seems the wrong `type`. 
Choice also represents irrelevant options such as "view submissions", "cancel dec". 

With this PR I propose we have a separate `DeclarationType`

What this PR changes:
- Replace `choice: String` on `ExportsDeclaration` with `type: DeclarationType`
- Use an Enum for DeclarationType which only allows `STANDARD`, `SUPPLEMENTARY` (for now)


app/uk/gov/hmrc/exports/models/DeclarationType.scala
```diff
@@ -0,0 +1,25 @@
...
+
+package uk.gov.hmrc.exports.models
+import play.api.libs.json._
+import uk.gov.hmrc.exports.util.EnumJson
+
+object DeclarationType extends Enumeration {
+  type DeclarationType = Value
+  implicit val format: Format[DeclarationType.Value] = EnumJson.format(DeclarationType)
+  val STANDARD, SUPPLEMENTARY, SIMPLIFIED = Value
+}
```
ed0906: Highlighting a Key change

ed0906: Wups I left Simplified in, oh well... saves another PR

PR #115, opened 2019-10-22
CEDS-1839 Permit Simplified Declarations

ed0906:
This permits Simplified Declarations following the same rules as Standard Declarations. These rules may change in subsequent stories.

As part of this change I have refactored AdditionalDeclarationType into an Enum, given there will be more of these to come I think now is a good opportunity to do it.

### DIFF
--- a/app/uk/gov/hmrc/exports/models/declaration/AdditionalDeclarationType.scala
+++ b/app/uk/gov/hmrc/exports/models/declaration/AdditionalDeclarationType.scala
@@ -16,15 +16,21 @@
 
 package uk.gov.hmrc.exports.models.declaration
 
-import play.api.libs.json.{Json, OFormat}
+import play.api.libs.json._
 
-case class AdditionalDeclarationType(additionalDeclarationType: String)
-
-object AdditionalDeclarationType {
-  implicit val format: OFormat[AdditionalDeclarationType] = Json.format[AdditionalDeclarationType]
-
-  object AllowedAdditionalDeclarationTypes {
-    val Simplified = "Y"
-    val Standard = "Z"
-  }
+object AdditionalDeclarationType extends Enumeration {
+  type AdditionalDeclarationType = Value
+  implicit val format: Format[AdditionalDeclarationType.Value] =
+    Format(
+      Reads(
+        _.validate[String]
+          .filter(s => AdditionalDeclarationType.values.exists(_.toString == s))
+          .map(s => AdditionalDeclarationType.values.find(_.toString == s).get)
+      ),
+      Writes(v => JsString(v.toString))
+    )
+  val SUPPLEMENTARY_SIMPLIFIED = Value("Y")
+  val SUPPLEMENTARY_EIDR = Value("Z")
+  val STANDARD_PRE_LODGED = Value("D")
+  val STANDARD_FRONTIER = Value("A")
 }
--- a/app/uk/gov/hmrc/exports/models/declaration/AdditionalDeclarationType.scala
+++ b/app/uk/gov/hmrc/exports/models/declaration/AdditionalDeclarationType.scala
@@ -24,8 +24,8 @@ object AdditionalDeclarationType extends Enumeration {
     Format(
       Reads(
         _.validate[String]
-          .filter(s => AdditionalDeclarationType.values.exists(_.toString == s))
-          .map(s => AdditionalDeclarationType.values.find(_.toString == s).get)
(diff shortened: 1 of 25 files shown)


app/uk/gov/hmrc/exports/models/declaration/AdditionalDeclarationType.scala
```diff
@@ -16,15 +16,21 @@
...
-    val Standard = "Z"
-  }
+object AdditionalDeclarationType extends Enumeration {
+  type AdditionalDeclarationType = Value
+  implicit val format: Format[AdditionalDeclarationType.Value] =
+    Format(
+      Reads(
+        _.validate[String]
+          .filter(s => AdditionalDeclarationType.values.exists(_.toString == s))
+          .map(s => AdditionalDeclarationType.values.find(_.toString == s).get)
```
PatrykRudnicki: Hmm, I think that will be better with just `find` without checking `exists` and getOrElse with exception.

The same proposal like on frontend about using `s` and `v`

app/uk/gov/hmrc/exports/services/mapping/declaration/SpecificCircumstancesCodeBuilder.scala
```diff
@@ -25,11 +25,13 @@ import wco.datamodel.wco.declaration_ds.dms._2._
...
-        .filter(_.circumstancesCode.contains("Yes"))
-        .map(_ => createCircumstancesCode)
-        .foreach(declaration.setSpecificCircumstancesCodeCode)
+    model.`type` match {
+      case DeclarationType.STANDARD | DeclarationType.SIMPLIFIED =>
+        model.locations.officeOfExit
+          .filter(_.circumstancesCode.contains("Yes"))
+          .map(_ => createCircumstancesCode)
+          .foreach(declaration.setSpecificCircumstancesCodeCode)
+      case _ =>
```
PatrykRudnicki: Can you add at least `(): Unit`. Empty case looks weird

app/uk/gov/hmrc/exports/services/mapping/declaration/consignment/DeclarationConsignmentBuilder.scala
```diff
@@ -29,11 +29,13 @@ class DeclarationConsignmentBuilder @Inject()(
...
-      consignmentCarrierBuilder.buildThenAdd(model, consignment)
-      declaration.setConsignment(consignment)
+    model.`type` match {
+      case DeclarationType.STANDARD | DeclarationType.SIMPLIFIED =>
+        val consignment = new Declaration.Consignment()
+        freightBuilder.buildThenAdd(model, consignment)
+        iteneraryBuilder.buildThenAdd(model, consignment)
+        consignmentCarrierBuilder.buildThenAdd(model, consignment)
+        declaration.setConsignment(consignment)
+      case _ =>
```
PatrykRudnicki: The same here

test/unit/uk/gov/hmrc/exports/services/mapping/goodsshipment/consignment/ConsignmentBuilderSpec.scala
```diff
@@ -27,24 +27,28 @@ import util.testdata.ExportsDeclarationBuilder
...
+  private val transportEquipmentBuilder = mock[TransportEquipmentBuilder]
 
-  private val builder =
-    new ConsignmentBuilder(mockGoodsLocationBuilder, mockContainerCodeBuilder, mockDepartureTransportMeansBuilder, mockTransportEquipmentBuilder)
+  private val builder = new ConsignmentBuilder(goodsLocationBuilder, containerCodeBuilder, departureTransportMeansBuilder, transportEquipmentBuilder)
+
+  override protected def beforeEach(): Unit = {
+    super.beforeEach()
+    reset(containerCodeBuilder, goodsLocationBuilder, departureTransportMeansBuilder, transportEquipmentBuilder)
+  }
```
PatrykRudnicki: I think this should be in `afterEach`. Resetting should be after test. Of course if you run it it doesn't make any difference, but IMO this is more logically if I use something and after this I'm resetting :)

app/uk/gov/hmrc/exports/models/declaration/AdditionalDeclarationType.scala
ed0906: Hmmm, i'd rather we get a JsError if we send something unexpected, it will make it obvious if we've missed something.

I'll change the `s` and `v`

PR #118, opened 2019-10-31
CEDS-1851 Adjust DMS mapping for implified Journey

PauloPortugal:
As part of this change we should not map Commodity Measurements and
Nature of Transaction, as they are not needed for the Simplified journey

### DIFF
--- a/app/uk/gov/hmrc/exports/services/mapping/goodsshipment/GoodsShipmentBuilder.scala
+++ b/app/uk/gov/hmrc/exports/services/mapping/goodsshipment/GoodsShipmentBuilder.scala
@@ -17,6 +17,7 @@
 package uk.gov.hmrc.exports.services.mapping.goodsshipment
 
 import javax.inject.Inject
+import uk.gov.hmrc.exports.models.DeclarationType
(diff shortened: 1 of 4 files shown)


app/uk/gov/hmrc/exports/services/mapping/goodsshipment/GoodsShipmentBuilder.scala
```diff
@@ -40,7 +41,9 @@ class GoodsShipmentBuilder @Inject()(
   override def buildThenAdd(exportsCacheModel: ExportsDeclaration, declaration: Declaration): Unit = {
     val goodsShipment = new GoodsShipment()
 
-    exportsCacheModel.natureOfTransaction.foreach(goodsShipmentNatureOfTransactionBuilder.buildThenAdd(_, goodsShipment))
+    if (exportsCacheModel.`type` != DeclarationType.SIMPLIFIED) {
```
PatrykRudnicki: I think it'll be more readable if we will check what journey can have those informations. I know the if will be longer but it will explain more if we will check Standard and Supplemetary instead of checking different than Simplified

PauloPortugal:
test this please

PR #120, opened 2019-11-04
CEDS-1720 Mapping for DMSREJ error messages

PatrykRudnicki:


PatrykRudnicki:
I can rename it but in separate PR, I don't know how long it'll take

PR #122, opened 2019-11-05
CEDS-1883 Enable containers for supplementary journey

PatrykRudnicki:


app/uk/gov/hmrc/exports/services/mapping/goodsshipment/consignment/ConsignmentBuilder.scala
```diff
@@ -43,7 +43,7 @@ class ConsignmentBuilder @Inject()(
     )
 
     exportsCacheModel.`type` match {
-      case DeclarationType.STANDARD | DeclarationType.SIMPLIFIED =>
+      case DeclarationType.STANDARD | DeclarationType.SIMPLIFIED | DeclarationType.SUPPLEMENTARY =>
```
PatrykRudnicki: Probably for now we have containers in all dec types. Do we want to stay with this case? or should I remove it?

SteveSugden: I was wondering tht when I did the front-end

PatrykRudnicki: I think it'll be good to stay with match here, still. I talked with @bambuchaAdm he agree with it.

PR #127, opened 2019-11-11
CEDS-1911 CUS code

SteveSugden:


app/uk/gov/hmrc/exports/services/mapping/governmentagencygoodsitem/GovernmentAgencyGoodsItemBuilder.scala
```diff
@@ -73,8 +73,9 @@ class GovernmentAgencyGoodsItemBuilder @Inject()(
   }
   private def mapItemTypeToCommodity(exportItem: ExportItem): Option[Commodity] =
     (exportItem.itemType, exportItem.commodityDetails) match {
-      case (Some(item), Some(details)) => Some(cachingMappingHelper.commodityFromItemTypes(item, details, exportItem.dangerousGoodsCode))
-      case _                           => None
+      case (Some(item), Some(details)) =>
+        Some(cachingMappingHelper.commodityFromItemTypes(item, details, exportItem.dangerousGoodsCode, exportItem.cusCode))
+      case _ => None
```
PatrykRudnicki: We're checking if `item` and `details` exists to return something but `dangerousGoodsCode` and `cusCode` are optional too. Should we validate it in the same way?

SteveSugden: This code has evolved slowly as I've been migrating fields.   dangerousGoodsCode and cusCode are were both optional fields on item and now they are optional pages containing an optional field.  (All pages are 'optional' according to our data model).  I'm fairly sure that given the same combination of input data this modified mapping code will produce the same results.  But maybe it could do with being re-factored.  I've still got two more pages to extract so I'll have another look when I do the last one

app/uk/gov/hmrc/exports/models/declaration/ExportItem.scala
```diff
@@ -139,6 +144,7 @@ case class ExportItem(
   itemType: Option[ItemType] = None,
   commodityDetails: Option[CommodityDetails] = None,
   dangerousGoodsCode: Option[UNDangerousGoodsCode] = None,
+  cusCode: Option[CUSCode] = None,
```
ed0906: could cusCode not just be an `Option[String]`?

PR #128, opened 2019-11-12
CEDS-1950 Make WarehouseIdentification backwards compatible

PauloPortugal:
As part of the story we have removed the `WarehouseIdentificationType`
from the user journey, but it is still required to go into the DMS
payload.

James W. and Joe will create a story to update the `hint` on the
warehouse IdentificationNumber page and to provide new validation rules
for the Warehouse type.
That ticket will be played in the following sprint.

In the meantime this change is simply not to downgrade the payload that
is sent to DMS, as DMS expects both identificationNumber and Type or
none of them.

The current Warehouse model will also be replaced in a subsequent story this sprint.

### DIFF
--- a/app/uk/gov/hmrc/exports/services/mapping/goodsshipment/WarehouseBuilder.scala
+++ b/app/uk/gov/hmrc/exports/services/mapping/goodsshipment/WarehouseBuilder.scala
@@ -30,17 +30,20 @@ class WarehouseBuilder @Inject()() extends ModifyingBuilder[WarehouseIdentificat
     }
 
   private def isDefined(warehouse: WarehouseIdentification): Boolean =
-    warehouse.identificationNumber.isDefined && warehouse.identificationType.isDefined
+    warehouse.identificationNumber.isDefined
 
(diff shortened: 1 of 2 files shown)


app/uk/gov/hmrc/exports/services/mapping/goodsshipment/WarehouseBuilder.scala
```diff
@@ -30,17 +30,20 @@ class WarehouseBuilder @Inject()() extends ModifyingBuilder[WarehouseIdentificat
     }
 
   private def isDefined(warehouse: WarehouseIdentification): Boolean =
-    warehouse.identificationNumber.isDefined && warehouse.identificationType.isDefined
+    warehouse.identificationNumber.isDefined
 
   private def createWarehouse(data: WarehouseIdentification): Warehouse = {
     val warehouse = new Warehouse()
 
+    val warehouseIdentificationType = data.identificationNumber.map(_.substring(0,1)).getOrElse("")
```
PatrykRudnicki: Hmm, I'm just wondering if taking `headOption` and `tail` will not be enough.

You're just extracting one character (first one)

PR #130, opened 2019-11-13
CEDS-1907 Change model for destination countries

PatrykRudnicki:


SteveSugden:
What happened to the DestinationCountries case class?  Is it still needed?

PatrykRudnicki:
Thanks @SteveSugden, Removed in next commit

PR #131, opened 2019-11-13
CEDS-1972 - fix for bug mapping CUS code to 'Classification'

SteveSugden:
Bug resulted in the creation of a Classification with a None identifier
(which is not permitted).
Took the opportunity to refactor the CachingMappingHelper to reduce the
expanding list of arguments passed to the
'getClassificationsFromItemTypes' method.

### DIFF
--- a/app/uk/gov/hmrc/exports/services/mapping/governmentagencygoodsitem/GovernmentAgencyGoodsItemBuilder.scala
+++ b/app/uk/gov/hmrc/exports/services/mapping/governmentagencygoodsitem/GovernmentAgencyGoodsItemBuilder.scala
@@ -64,28 +64,24 @@ class GovernmentAgencyGoodsItemBuilder @Inject()(
     wcoGovernmentAgencyGoodsItem: WCOGovernmentAgencyGoodsItem,
     declarationType: DeclarationType
   ): Unit = {
-    val combinedCommodity = for {
-      commodityWithoutGoodsMeasure <- mapExportItemToCommodity(exportItem)
-      commodityOnlyGoodsMeasure = mapCommodityMeasureToCommodity(exportItem.commodityMeasure, declarationType)
-    } yield combineCommodities(commodityWithoutGoodsMeasure, commodityOnlyGoodsMeasure)
+    val commodityWithoutGoodsMeasure = mapCommodityValuesToCommodity(exportItem)
+    val commodityOnlyGoodsMeasure = mapCommodityMeasureToCommodity(exportItem.commodityMeasure, declarationType)
 
-    combinedCommodity.foreach(commodityBuilder.buildThenAdd(_, wcoGovernmentAgencyGoodsItem))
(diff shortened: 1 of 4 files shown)


app/uk/gov/hmrc/exports/services/mapping/governmentagencygoodsitem/GovernmentAgencyGoodsItemBuilder.scala
```diff
@@ -63,19 +63,19 @@ class GovernmentAgencyGoodsItemBuilder @Inject()(
...
-    (exportItem.itemType, exportItem.commodityDetails) match {
-      case (Some(item), Some(details)) =>
-        Some(cachingMappingHelper.commodityFromItemTypes(item, details, exportItem.dangerousGoodsCode, exportItem.cusCode, exportItem.taricCodes))
-      case _ => None
+  private def mapExportItemToCommodity(exportItem: ExportItem): Option[Commodity] =
+    if (exportItem.commodityDetails.isDefined) {
+      Some(cachingMappingHelper.commodityFromExportItem(exportItem))
+    } else {
+      None
     }
```
SteveSugden: This method used to be:
```
  private def mapItemTypeToCommodity(itemType: Option[ItemType]): Option[Commodity] =
    itemType.map({ item =>
      cachingMappingHelper.commodityFromItemTypes(item)
    })
```

As `ItemType` is/was a required page, the old method will always return Some(Commodity)

The new version does the same thing with `CommodityDetails` page - which is also a page that is always shown - so the effect is the same.

app/uk/gov/hmrc/exports/services/mapping/governmentagencygoodsitem/GovernmentAgencyGoodsItemBuilder.scala
```diff
@@ -63,20 +63,16 @@ class GovernmentAgencyGoodsItemBuilder @Inject()(
...
 
     combinedCommodity.foreach(commodityBuilder.buildThenAdd(_, wcoGovernmentAgencyGoodsItem))
   }
-  private def mapItemTypeToCommodity(exportItem: ExportItem): Option[Commodity] =
-    (exportItem.itemType, exportItem.commodityDetails) match {
-      case (Some(item), Some(details)) =>
-        Some(cachingMappingHelper.commodityFromItemTypes(item, details, exportItem.dangerousGoodsCode, exportItem.cusCode, exportItem.taricCodes))
-      case _ => None
-    }
+  private def mapExportItemToCommodity(exportItem: ExportItem): Option[Commodity] =
```
SteveSugden: This method used to be:
```
  private def mapItemTypeToCommodity(itemType: Option[ItemType]): Option[Commodity] =
    itemType.map({ item =>
      cachingMappingHelper.commodityFromItemTypes(item)
    })
```

As `ItemType` is/was a required page, so the old method always returned Some(Commodity) which then had to be handled using the for comprehension.

PR #132, opened 2019-11-14
CEDS 1914 NACT code

SteveSugden:


app/uk/gov/hmrc/exports/models/declaration/ExportItem.scala
```diff
@@ -63,7 +63,12 @@ object TaricCode {
   implicit val format: OFormat[TaricCode] = Json.format[TaricCode]
 }
 
-case class ItemType(nationalAdditionalCode: Seq[String], statisticalValue: String)
+case class NactCode(nactCode: String)
```
bambuchaAdm: Let use full name would be less confusing for new people

PatrykRudnicki: I prefer full name `NationalAdditionalCode`.

PatrykRudnicki: Ok it's name from ticket

SteveSugden: Its all confusing.  Whats a "TARIC" or a CUS

SteveSugden: Was just about to point that out - name matches the UI

PR #137, opened 2019-11-25
CEDS 1957 - extract payment and has containers from border transport

SteveSugden:
This PR extracts both the "transport payment" and "has containers" question from the "border transport".   In addition, the contain data is re-located into the new "TransportData" top level of the cache model.


app/uk/gov/hmrc/exports/controllers/request/ExportsDeclarationRequest.scala
```diff
@@ -34,7 +34,7 @@ case class ExportsDeclarationRequest(
   consignmentReferences: Option[ConsignmentReferences] = None,
   departureTransport: Option[DepartureTransport] = None,
   borderTransport: Option[BorderTransport] = None,
-  containerData: Option[TransportInformationContainers] = None,
+  transportData: Option[TransportData] = None,
```
ed0906: can we not suffix it with `data`

SteveSugden: Are you saying you'd prefer not to have the suffix?

PR #140, opened 2019-12-11
Ceds 2036 Accept two declaration shapes at one time

bambuchaAdm:


ed0906:
Can you add a summary with why you're doing this?

test/unit/uk/gov/hmrc/exports/models/declaration/AdditionalDeclarationTypeSpec.scala
```diff
@@ -14,10 +14,11 @@
  * limitations under the License.
  */
 
-package uk.gov.hmrc.exports.models.declaration
+package unit.uk.gov.hmrc.exports.models.declaration
```
ed0906: Why?

ed0906:
I'm not convinced for the need for this. We are in a fortunate position where we are not live, and have no real data of value to protect. So while we can we should make the most of it and keep the model simple. One version. If the Service breaks for 30s during builds then so be it :)

bambuchaAdm:
> We are in a fortunate position where we are not live, and have no real data of value to protect.

This is thing we have to build in some point of time - if there would be one reads/writes pair is same > as having it generated using macros. We have to pay this - now or later. Hopefully this is one time transaction.

> If the Service breaks for 30s during builds then so be it :)
It breaks microservice principle about having independent deployment units.

ed0906:
> It breaks microservice principle about having independent deployment units.

They are still very much independent. Just in the rare instance we need to change the model the build might break for 30-60s. In my eyes thats a suitable cost to keep our technical debt down. Adding stuff like this (when we ultimately don't need it) means we later have to remove it again, and I guarantee the cost of removal is far greater than a temporary build break (Create JIRA ticket, prioritise, remove the code, create a PR, wait for it to be approved, wait for the build to go green)

bambuchaAdm:
After discussion. I don't assume to have this one longer then scope of this ticket. I would remove version1 just after frontend changes would be applied. 

I made those version mapping to avoid to change logic inside builders that would handle transision shape of declaration.

@ed0906 
>(Create JIRA ticket, prioritise, remove the code, create a PR, wait for it to be approved, wait for the build to go green)
All those changes would be made in scope of single ticket. There could be multiple PR but is same case as there would be changes from QA phase. 

Discussion convinced me that validation on B/E is not good idea now.

PR #142, opened 2019-12-16
CEDS-2036 Better mapping in models

bambuchaAdm:


SteveSugden:
What's the impact of changing this logic?   We'll now always return Some.  Or was that always the case (i.e. the None branches were never used)?

PatrykRudnicki:
@SteveSugden This logic has been implemented with adding the possibility to handle two versions of the declaration. The implementation was wrong, this is fix for it. I think the title don't describe what has been done here.

PR #143, opened 2019-12-16
Update dependencies

PatrykRudnicki:


app/uk/gov/hmrc/exports/repositories/SubmissionRepository.scala
```diff
@@ -93,11 +92,9 @@ class SubmissionRepository @Inject()(implicit mc: ReactiveMongoComponent, ec: Ex
...
   private def performUpdate(query: JsObject, update: JsObject): Future[Option[Submission]] =
     findAndUpdate(query, update, fetchNewObject = true).map { updateResult =>
-      if (updateResult.value.isEmpty) logDatabaseUpdateError(updateResult)
+      if (updateResult.value.isEmpty) {
+        updateResult.lastError.foreach(_.err.foreach(errorMsg => logger.error(s"Problem during database update: $errorMsg")))
+      }
       updateResult.result[Submission]
     }
-
-  private def logDatabaseUpdateError(res: FindAndModifyResult): Unit =
```
PatrykRudnicki: FindAndModifyResult is depracated and in next versions will be unavailable to use. There was some problem with import. This method has been used in just one place so inline solved the problem.

project/plugins.sbt
```diff
@@ -21,3 +21,5 @@ addSbtPlugin("org.scoverage" % "sbt-scoverage" % "1.5.1")
 addSbtPlugin("com.typesafe.sbt" % "sbt-digest" % "1.1.2")
 
 addSbtPlugin("com.lucidchart" % "sbt-scalafmt" % "1.16")
+
+addSbtPlugin("uk.gov.hmrc" % "sbt-git-stamp" % "6.1.0")
```
PatrykRudnicki: We're using `gitStampSettings` in `build.sbt` with upgrades we need to specify this manually to have access to those settings. Not sure that we need it

PR #152, opened 2020-01-13
CEDS-2086 - only send eori (preferred) or address for parties

SteveSugden:


app/uk/gov/hmrc/exports/services/mapping/goodsshipment/consignment/ConsignmentCarrierBuilder.scala
```diff
@@ -45,27 +45,27 @@ class ConsignmentCarrierBuilder @Inject()(countriesService: CountriesService) ex
...
 
-    details.address
-      .filter(_.fullName.nonEmpty)
-      .foreach { address =>
-        val carrierName = new CarrierNameTextType()
-        carrierName.setValue(address.fullName)
-        carrier.setName(carrierName)
-      }
+        details.address
+          .filter(address => address.addressLine.nonEmpty || address.townOrCity.nonEmpty || address.postCode.nonEmpty || address.country.nonEmpty)
```
PatrykRudnicki: Probably it's worth to add method to the address to check if it's defined. It's not blocking comment

SteveSugden: That's a good idea.  That logic is used  >1 I think.   I merged as soon as I saw approved.  So I'll create a quick 2nd PR tomorrow

SteveSugden: OTOH - just looked at this PR and all the builders seem to handle the completeness of the address in a different way.  Some don't care if elements are missing; some only copy over completed elements, etc.  This builder seems to be the only one that only sets an address if all elements are present.

Given how "picky" the downstream service is about the request we send them, it might not be a good idea to change anything.  If it ain't broke....

PR #159, opened 2020-03-12
BAU - Remove mongobee and add mongock

PatrykRudnicki:


app/uk/gov/hmrc/exports/mongock/changesets/CacheChangelog.java
```diff
@@ -0,0 +1,14 @@
...
+
+import com.github.cloudyrock.mongock.ChangeLog;
+import com.github.cloudyrock.mongock.ChangeSet;
+import com.mongodb.client.MongoDatabase;
+
+@ChangeLog
+public class CacheChangelog {
+    private String collection = "declarations";
+
+    @ChangeSet(order = "001", id = "Deckaration DB Baseline", author = "Patryk Rudnicki")
```
PauloPortugal: you need to keep the same changeset info, otherwise it will apply it again:

```
@ChangeSet(order = "001", id = "Exports DB Baseline", author = "Paulo Monteiro")
```

PauloPortugal: But double check locally

PatrykRudnicki: I change it. Thanks for the review. I'll double check it locally.

PatrykRudnicki:
@SteveSugden You disabled this, not removed :)

PatrykRudnicki:
@SteveSugden @PauloPortugal I need review once again because of the new commit. I will squash them during merging

app/uk/gov/hmrc/exports/mongock/changesets/CacheChangelog.java
```diff
@@ -0,0 +1,14 @@
+package uk.gov.hmrc.exports.mongock.changesets;
+
+import com.github.cloudyrock.mongock.ChangeLog;
+import com.github.cloudyrock.mongock.ChangeSet;
+import com.mongodb.client.MongoDatabase;
+
+@ChangeLog
+public class CacheChangelog {
+    private String collection = "declarations";
```
MaciejRewera: What about other collections? (Notifications and Submissions)

PatrykRudnicki: We don't need for now to change those collections. If necessary we will just add it.

PR #161, opened 2020-03-16
CEDS-2171 Log the response body when an error occurred

PauloPortugal:


app/uk/gov/hmrc/exports/connectors/ead/CustomsDeclarationsInformationConnector.scala
```diff
@@ -44,6 +44,7 @@ class CustomsDeclarationsInformationConnector @Inject()(mrnStatusParser: MrnStat
             logger.debug(s"CUSTOMS_DECLARATIONS_INFORMATION fetch MRN status response ${response.body}")
             Some(mrnStatusParser.parse(xml.XML.loadString(response.body)))
           case status =>
+            logger.debug(s"CUSTOMS_DECLARATIONS_INFORMATION fetch MRN status response ${response.body}")
```
SteveSugden: Is there any point logging an error at debug level?  I presume that debug isn't switched on anywhere useful

PauloPortugal: Good point. I have changed it to `WARN`

bambuchaAdm: warn probably would be the best

PR #162, opened 2020-03-16
CEDS-2191 - Change validation on Commodity Measurements page for Clearance Requests

SteveSugden:
This PR makes gross and net weights on the CommodityMeasure optional.  I've tested that this BE change is compatible with the existing FE by creating a draft declaration with the existing BE then switching to this branch and completing the dec and submitting it.  (i.e. db upgrades are not necessary).

Only thing that I can't check is that we can eventually submit a clearance declaration without a CommodityMeasure.   It _should_ work because the wco `Commodity` that we create is optional (both in the wco library and the customs_declaration XSD).  But that will need to be verified once we have completed and deployed the FE changes for clearance requests to QA

### DIFF
--- a/app/uk/gov/hmrc/exports/services/mapping/CachingMappingHelper.scala
+++ b/app/uk/gov/hmrc/exports/services/mapping/CachingMappingHelper.scala
@@ -51,11 +51,16 @@ class CachingMappingHelper {
         taricCodes.map(code => Classification(Some(code), identificationTypeCode = Some(TARICAdditionalCode.value)))
     ).toSeq
 
-  def mapGoodsMeasure(data: CommodityMeasure) =
-    Commodity(
-      goodsMeasure =
-        Some(GoodsMeasure(Some(createMeasure(data.grossMass)), Some(createMeasure(data.netMass)), data.supplementaryUnits.map(createMeasure)))
-    )
+  def mapGoodsMeasure(data: CommodityMeasure): Option[Commodity] = data match {
+    case CommodityMeasure(None, None, None) => None
+    case _ =>
+      Some(
+        Commodity(
+          goodsMeasure =
+            Some(GoodsMeasure(data.grossMass.map(createMeasure), data.netMass.map(createMeasure), data.supplementaryUnits.map(createMeasure)))
+        )
+      )
+  }
 
   private def createMeasure(unitValue: String) =
     try {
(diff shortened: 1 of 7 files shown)


app/uk/gov/hmrc/exports/services/mapping/CachingMappingHelper.scala
```diff
@@ -51,11 +51,16 @@ class CachingMappingHelper {
...
     ).toSeq
 
-  def mapGoodsMeasure(data: CommodityMeasure) =
-    Commodity(
-      goodsMeasure =
-        Some(GoodsMeasure(Some(createMeasure(data.grossMass)), Some(createMeasure(data.netMass)), data.supplementaryUnits.map(createMeasure)))
-    )
+  def mapGoodsMeasure(data: CommodityMeasure): Option[Commodity] = data match {
+    case CommodityMeasure(None, None, None) => None
+    case _ =>
```
PatrykRudnicki: My propose is to change it to below. To extract values during pattern matching. 

```
case CommodityMeasure(supplementaryUnits, netMass, grossMass) =>
  Some(
    Commodity(
          goodsMeasure =
            Some(GoodsMeasure(grossMass.map(createMeasure), netMass.map(createMeasure), supplementaryUnits.map(createMeasure)))
        )
  )
```

SteveSugden: I thought of doing that but couldn't see any real advantage.  But in the spirit of the times I will update.

PatrykRudnicki: Using pattern matching just to check if all the values are empty and after that come back to using the element that you're matching is not a good idea. For this approach is better to check something with `if` or prepare a method inside the `CommodityMeasure` class to check if it's empty.

SteveSugden: I've done it.   But I originally did it wrong.  I matched 
`case CommodityMeasure(grossMass, netMass, supplementaryUnits)`
i.e. I defined the variables in the order I was later using them.   It compiled (of course as they are all the same type).  Only spotted the error when a test failed.

So on reflection, my original way was less error prone.  IMO.

PatrykRudnicki: Agree with you @SteveSugden but still I think it's not the reason of the using pattern matching :)

PR #165, opened 2020-03-17
CEDS-2231 Add migration for location country and clear Goods Location model

PatrykRudnicki:


app/uk/gov/hmrc/exports/services/mapping/goodsshipment/consignment/GoodsLocationBuilder.scala
```diff
@@ -71,27 +61,9 @@ class GoodsLocationBuilder @Inject()(countriesService: CountriesService) extends
...
-    }
-
-    goods.postCode.foreach { value =>
-      val postcode = new AddressPostcodeIDType()
-      postcode.setValue(value)
-      goodsAddress.setPostcodeID(postcode)
-    }
-
     if (goods.country.nonEmpty) {
       val countryCode = new AddressCountryCodeType
```
SteveSugden: So we were passing the country name and then looking up the code in the backend to pass to xml?

PatrykRudnicki: Nope, we're passing country code from the frontend. This is the reason that I'm not manipulating data here.

Previously we had country as dropdown, actually it's one input with code e.g. `GBAUFXTFXTFXT` so we're passing two first characters to the backend and just put this into the payload.

app/uk/gov/hmrc/exports/mongock/changesets/CacheChangelog.java
```diff
@@ -27,4 +33,27 @@
...
     }
+
+    @ChangeSet(order = "002", id = "CEDS-2231 Change country name to country code for location page", author = "Patryk Rudnicki")
+    public void updateAllCountriesNameToCodesForLocationPage(MongoDatabase db) {
+        Document query = new Document();
+        CountriesService service = new CountriesService();
+
+        FindIterable<Document> documents = db.getCollection(collection).find(new BasicDBObject(query));
+
+        for (Document document : documents) {
```
SteveSugden: Just to check I'm reading this right.    
You are looking for existing declarations with a country name in location>goodsLocation>country converting that country name to code and reseting it back on the same country field?

app/uk/gov/hmrc/exports/mongock/MongockConfig.scala
```diff
@@ -27,7 +27,7 @@ case class MongockConfig(mongoURI: String) {
 
   val client = new MongoClient(uri)
 
-  val runner = new MongockBuilder(client, uri.getDatabase, "mongock.changesets")
+  val runner = new MongockBuilder(client, uri.getDatabase, "uk.gov.hmrc.exports.mongock.changesets")
```
SteveSugden: Was the full package name necessary or is this just to be more "accurate".

app/uk/gov/hmrc/exports/services/mapping/goodsshipment/consignment/GoodsLocationBuilder.scala
SteveSugden: Yes.  I was just commented on what we used to do.  I couldn't find the facepalm icon!

app/uk/gov/hmrc/exports/mongock/changesets/CacheChangelog.java
PatrykRudnicki: 1) I'm going through all existing declarations
2) Checking if country field is defined
3) If it's defined I'm retrieving existing country name
4) Based on the country name I'm generating country code
5) Replacing old value in the model from country name to country code
6) Replacing whole declaration with updated on inside the cache

app/uk/gov/hmrc/exports/mongock/MongockConfig.scala
PatrykRudnicki: It's necessary, because it's backend service, so backend service has this as the package.

In the frontend application it wasn't necessary, because we don't have packages but directly we're putting stuff inside the `app` folder.

Previous implementation didn't work. I didn't check it (my fault) just check if the application works correctly

PatrykRudnicki:
@SteveSugden yes, I'm planning to prepare the data on QA and see if this changed correctly.

I tested everything locally, if you're interested in the results I attached developers note to the ticket and provided result of local testing

PR #167, opened 2020-03-18
CEDS-2231 Protect countries code to be replaced by empty strings during migration

PatrykRudnicki:


app/uk/gov/hmrc/exports/services/CountriesService.scala
```diff
@@ -41,7 +41,8 @@ class CountriesService @Inject()() {
 
   private def countryCode: String => String = cc => cc.split(":")(1).trim
 
-  def findCountryCode(countryName: String): String = allCountries.find(_.countryName == countryName).map(_.countryCode).getOrElse("")
+  def findCountryCodeOrReturnCountryName(countryName: String): String =
+    allCountries.find(_.countryName == countryName).map(_.countryCode).getOrElse(countryName)
```
MaciejRewera: Wouldn't be better to return `Option[String]` and let the client decide what to do in case of not finding the country code?

PatrykRudnicki: It is used in Java class. The normal `getOrElse` doesn't work there

PatrykRudnicki: I tried to prepare specific method in Java, but with my knowledge it's difficult to use scala list with countries (case class Country) to do this inside the Java class

PatrykRudnicki: If you have any ideas please let me know :) I don't like this solution, but we need to make something to protect the data

SteveSugden: Could go old school and return null!  Or how about the client provides a default value if no mapping found?

app/uk/gov/hmrc/exports/mongock/changesets/ScalaCacheChangeLog.scala
```diff
@@ -0,0 +1,71 @@
...
+      val objectToBeUpdated: BasicDBObject = new BasicDBObject(queryIndexes)
+
+      db.getCollection(collection).replaceOne(objectToBeUpdated, document)
+    }
+  }
+
+  private def findCountryCode(countryName: String): String = {
+    val service: CountriesService = new CountriesService
+
+    service.allCountries.find(_.countryName == countryName).map(_.countryCode).getOrElse(countryName)
```
SteveSugden: 👍 So much easier when you're in scala-land.  I might be a convert after all

app/uk/gov/hmrc/exports/mongock/changesets/ScalaCacheChangeLog.scala
```diff
@@ -0,0 +1,71 @@
...
+
+import com.github.cloudyrock.mongock.{ChangeLog, ChangeSet}
+import com.google.common.collect.ImmutableMap
+import com.mongodb.BasicDBObject
+import com.mongodb.client.MongoDatabase
+import org.bson.Document
+import uk.gov.hmrc.exports.services.CountriesService
+
+@ChangeLog
+class ScalaCacheChangeLog {
```
MaciejRewera: Might be worth removing "Scala" from the name (?)

app/uk/gov/hmrc/exports/mongock/changesets/ScalaCacheChangeLog.scala
PatrykRudnicki: Agree!

app/uk/gov/hmrc/exports/mongock/changesets/ScalaCacheChangeLog.scala
PatrykRudnicki: Good point

PatrykRudnicki: Changed

app/uk/gov/hmrc/exports/mongock/changesets/ScalaCacheChangeLog.scala
```diff
@@ -0,0 +1,71 @@
...
+      val queryIndexes: Map[String, String] =
+        ImmutableMap.of("id", document.get("id").asInstanceOf[String], "eori", document.get("eori").asInstanceOf[String])
+
+      val objectToBeUpdated: BasicDBObject = new BasicDBObject(queryIndexes)
+
+      db.getCollection(collection).replaceOne(objectToBeUpdated, document)
+    }
+  }
+
+  private def findCountryCode(countryName: String): String = {
```
MaciejRewera: 👍

MaciejRewera: Might need to create a separate object with those helper methods. Not to pollute this class with too many helpers.

PatrykRudnicki: I think it's the best to keep this close to the script, because it can be use just for this

MaciejRewera: I meant a helper object/class for `CacheChangeLog` only. Best, in the same package.

PR #169, opened 2020-03-19
CEDS-2189 Make CommodityDetails.descriptionOfGoods optional

PauloPortugal:
For Clearance Requests, data elements '6/14 Commodity code' and
'6/8 Description of Goods' are not required in the majority of cases.

This data element is not required on clearance requests using procedure
codes 00 18 or 00 19 in D.E. 1/10.

**Relates to https://github.com/hmrc/customs-declare-exports-frontend/pull/788**

### DIFF
--- a/app/uk/gov/hmrc/exports/services/mapping/CachingMappingHelper.scala
+++ b/app/uk/gov/hmrc/exports/services/mapping/CachingMappingHelper.scala
@@ -23,20 +23,25 @@ import uk.gov.hmrc.wco.dec._
 class CachingMappingHelper {
   val defaultMeasureCode = "KGM"
 
-  def commodityFromExportItem(exportItem: ExportItem): Commodity = {
-    def removeCarriageReturns(description: String): String = description.replaceAll("(\\r\\n)|\\r|\\n", " ")
+  def commodityFromExportItem(exportItem: ExportItem): Option[Commodity] = {
+    def removeCarriageReturns(description: Option[String]): Option[String] = description.map(_.replaceAll("(\\r\\n)|\\r|\\n", " "))
 
-    Commodity(
-      description = exportItem.commodityDetails.map(commodityDetails => removeCarriageReturns((commodityDetails.descriptionOfGoods))),
-      classifications = getClassifications(
-        exportItem.commodityDetails.flatMap(_.combinedNomenclatureCode),
-        exportItem.cusCode.flatMap(_.cusCode),
-        exportItem.nactCodes.map(_.map(_.nactCode)).getOrElse(List.empty),
-        exportItem.taricCodes.map(_.map(_.taricCode)).getOrElse(List.empty)
-      ),
-      dangerousGoods = exportItem.dangerousGoodsCode.flatMap(_.dangerousGoodsCode).map(code => Seq(DangerousGoods(Some(code)))).getOrElse(Seq.empty)
-    )
+    Some(
+      Commodity(
+        description = exportItem.commodityDetails.flatMap(commodityDetails => removeCarriageReturns(commodityDetails.descriptionOfGoods)),
+        classifications = getClassifications(
+          exportItem.commodityDetails.flatMap(_.combinedNomenclatureCode),
+          exportItem.cusCode.flatMap(_.cusCode),
+          exportItem.nactCodes.map(_.map(_.nactCode)).getOrElse(List.empty),
+          exportItem.taricCodes.map(_.map(_.taricCode)).getOrElse(List.empty)
+        ),
+        dangerousGoods = exportItem.dangerousGoodsCode.flatMap(_.dangerousGoodsCode).map(code => Seq(DangerousGoods(Some(code)))).getOrElse(Seq.empty)
+      )
+    ).filter(isCommodityNonEmpty)
   }
+
+  private def isCommodityNonEmpty(commodity: Commodity): Boolean =
+    commodity.classifications.nonEmpty || commodity.dangerousGoods.nonEmpty || commodity.description.nonEmpty
 
   private def getClassifications(
     commodityCode: Option[String],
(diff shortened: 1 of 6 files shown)


app/uk/gov/hmrc/exports/services/mapping/CachingMappingHelper.scala
```diff
@@ -24,10 +24,10 @@ class CachingMappingHelper {
   val defaultMeasureCode = "KGM"
 
   def commodityFromExportItem(exportItem: ExportItem): Commodity = {
-    def removeCarriageReturns(description: String): String = description.replaceAll("(\\r\\n)|\\r|\\n", " ")
+    def removeCarriageReturns(description: Option[String]): Option[String] = description.map(_.replaceAll("(\\r\\n)|\\r|\\n", " "))
 
     Commodity(
```
SteveSugden: Q.  If the resulting Commodity now ends up containing nothing (because all the inputs are now optional) do we need to change this method to return Option[Commodity] and ignore this part of in the XML payload. 
I did something similar recently when changing the CommodityMeasure code in the BE.

test/unit/uk/gov/hmrc/exports/services/mapping/CachingMappingHelperSpec.scala
```diff
@@ -104,6 +104,17 @@ class CachingMappingHelperSpec extends WordSpec with Matchers {
         commodity.classifications shouldBe Seq.empty
       }
 
+      "No commodity code or description provided" in {
+        val exportItem = ExportItem("id", commodityDetails = Some(CommodityDetails(None, None)))
+
+        val commodity = new CachingMappingHelper().commodityFromExportItem(exportItem)
```
SteveSugden: Here for example - should Commodity be 'None' in this case?

app/uk/gov/hmrc/exports/services/mapping/CachingMappingHelper.scala
PauloPortugal: Ahhh.... Really good point!

test/unit/uk/gov/hmrc/exports/services/mapping/CachingMappingHelperSpec.scala
PauloPortugal: Yes, I think it should be :) Thank you!

app/uk/gov/hmrc/exports/services/mapping/CachingMappingHelper.scala
PauloPortugal: This has been addressed!

test/unit/uk/gov/hmrc/exports/services/mapping/CachingMappingHelperSpec.scala
PauloPortugal: This has been addressed!

PauloPortugal:
@SteveSugden please review again :)

app/uk/gov/hmrc/exports/services/mapping/CachingMappingHelper.scala
```diff
@@ -17,24 +17,34 @@
...
-        exportItem.commodityDetails.flatMap(_.combinedNomenclatureCode),
-        exportItem.cusCode.flatMap(_.cusCode),
-        exportItem.nactCodes.map(_.map(_.nactCode)).getOrElse(List.empty),
-        exportItem.taricCodes.map(_.map(_.taricCode)).getOrElse(List.empty)
-      ),
-      dangerousGoods = exportItem.dangerousGoodsCode.flatMap(_.dangerousGoodsCode).map(code => Seq(DangerousGoods(Some(code)))).getOrElse(Seq.empty)
+    exportItem.commodityDetails.flatMap(
+      commodityDetails =>
+        commodityDetails match {
+          case CommodityDetails(None, None) => None
```
SteveSugden: What if CommodityDetails is (None, None) but there are/is a dangerousGoodsCode on the item.  It will be lost now won't it?

I would only return None (CommodityDetails) iff
- no descriptionOfGoods
- classifications is empty
- no dangerousGoodsCode

SteveSugden: Similar - you are discarding the CUS, NACT and TARIC codes at the moment

PauloPortugal: :man_facepalming: 
Thank you once again. 
I have used a private method to check if the object is "empty", instead of adding it to the wco-dec model. 
We will have to remove the wco-dec from this project, although it depends on it. :shrug:

app/uk/gov/hmrc/exports/services/mapping/CachingMappingHelper.scala
```diff
@@ -23,21 +23,26 @@ import uk.gov.hmrc.wco.dec._
...
+          exportItem.cusCode.flatMap(_.cusCode),
+          exportItem.nactCodes.map(_.map(_.nactCode)).getOrElse(List.empty),
+          exportItem.taricCodes.map(_.map(_.taricCode)).getOrElse(List.empty)
+        ),
+        dangerousGoods = exportItem.dangerousGoodsCode.flatMap(_.dangerousGoodsCode).map(code => Seq(DangerousGoods(Some(code)))).getOrElse(Seq.empty)
+      )
+    ).filter(isCommodityEmpty)
   }
 
+  private def isCommodityEmpty(commodity: Commodity): Boolean =
```
SteveSugden: One final suggestion.   Should this method name not be negated?  I.e. `isCommodityNonEmpty'?

And good point about not adding method to wco model.  Why will we have to remove the wco models BTW?

PauloPortugal: Yes it should be name `isCommodityNonEmpty` :+1: 
The Models were created in `wco-dec` library and we should really not used them, as agreed a few months back, but somehow these the item level models have not been migrated.

PauloPortugal:
Smoke tests were green.

PR #170, opened 2020-03-19
CEDS-2246 Make Declarant Details optional and remove non existing address

PatrykRudnicki:


app/uk/gov/hmrc/exports/services/mapping/declaration/DeclarantBuilder.scala
```diff
@@ -36,52 +35,16 @@ class DeclarantBuilder @Inject()(countriesService: CountriesService) extends Mod
...
-
-  private def deriveCountryCode(addressCountry: String): String =
-    countriesService.allCountries
-      .find(country => addressCountry.contains(country.countryName))
-      .map(_.countryCode)
-      .getOrElse("")
-
-  private def isDefined(declarantDetails: DeclarantDetails): Boolean =
-    declarantDetails.details.eori.isDefined || declarantDetails.details.address.isDefined
+  private def isDefined(declarantDetails: DeclarantDetails): Boolean = declarantDetails.details.flatMap(_.eori).isDefined
```
MaciejRewera: Is it used anywhere now?

app/uk/gov/hmrc/exports/models/declaration/Parties.scala
```diff
@@ -19,31 +19,37 @@ package uk.gov.hmrc.exports.models.declaration
...
 }
 
 case class ConsigneeDetails(details: EntityDetails)
+
 object ConsigneeDetails {
   implicit val format: OFormat[ConsigneeDetails] = Json.format[ConsigneeDetails]
 }
 
-case class DeclarantDetails(details: EntityDetails)
+case class DeclarantDetails(details: Option[EntityDetails])
```
MaciejRewera: Do we have a scenario when we have no declarant details? I thought this is mandatory for the whole declaration, no matter what type it has. How would we be able to tell who has submitted it?

app/uk/gov/hmrc/exports/services/mapping/declaration/DeclarantBuilder.scala
PatrykRudnicki: Yes, in the first method in this class

app/uk/gov/hmrc/exports/models/declaration/Parties.scala
PatrykRudnicki: DeclarantDetails are optional in Parties. This element is mandatory in every journey except Clearance

MaciejRewera: Besides, in the `ExportsDeclaration` model this field is already optional, so why would we need to have so many nested `Option`s ?

test/unit/uk/gov/hmrc/exports/services/mapping/declaration/DeclarantBuilderSpec.scala
```diff
@@ -16,75 +16,57 @@
...
-        declaration.getDeclarant should be(null)
+        declaration.getDeclarant shouldBe null
       }
 
-      "no eori" in {
-        val model = aDeclaration(withDeclarantDetails(eori = None, address = Some(Address("name", "line", "city", "postcode", "United Kingdom"))))
-        val declaration = new Declaration()
-
-        builder.buildThenAdd(model, declaration)
+      "empty declarant details" in {
```
MaciejRewera: This test case and the next one are exactly the same.

test/unit/uk/gov/hmrc/exports/services/mapping/declaration/DeclarantBuilderSpec.scala
```diff
@@ -16,75 +16,57 @@
...
         val declaration = new Declaration()
 
         builder.buildThenAdd(model, declaration)
 
-        declaration.getDeclarant.getID should be(null)
-        declaration.getDeclarant.getAddress.getCountryCode.getValue should be("")
+        declaration.getDeclarant shouldBe null
       }
 
       "populated" in {
```
MaciejRewera: Nit: What do you think about changing this bit? To me, "populated" is more suitable for views.

test/unit/uk/gov/hmrc/exports/services/mapping/declaration/DeclarantBuilderSpec.scala
PatrykRudnicki: empty declarant details means empty but visited. The model is different. Behaviour the same

test/unit/uk/gov/hmrc/exports/services/mapping/declaration/DeclarantBuilderSpec.scala
PatrykRudnicki: Hmm, not sure. Any proposals?

app/uk/gov/hmrc/exports/models/declaration/Parties.scala
PatrykRudnicki: That's a good question. As I understand this approach allow us to present empty element on the summary page. The same approach is on the representative details

PatrykRudnicki:
Closing for now, probably unnecessary change

test/unit/uk/gov/hmrc/exports/services/mapping/declaration/DeclarantBuilderSpec.scala
MaciejRewera: Ok I see

PR #171, opened 2020-03-23
CEDS-2192 - ignore in-applicable means of transport on departure type

SteveSugden:
This is the companion PR to https://github.com/hmrc/customs-declare-exports-frontend/pull/793
It simply discards the means of transport on departure information if the user selected the "not applicable" or none option when building the XML payload.

### DIFF
--- a/app/uk/gov/hmrc/exports/models/declaration/Transport.scala
+++ b/app/uk/gov/hmrc/exports/models/declaration/Transport.scala
@@ -38,11 +38,12 @@ case class Transport(
 
   def hasDepartureTransportDetails: Boolean =
     meansOfTransportOnDepartureIDNumber.nonEmpty || meansOfTransportOnDepartureType.nonEmpty
+
+  def isMeansOfTransportOnDepartureDefined: Boolean = meansOfTransportOnDepartureType.exists(_ != Transport.optionNone)
 }
 
 object Transport {
   implicit val format: OFormat[Transport] = Json.format[Transport]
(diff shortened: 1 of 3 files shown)


app/uk/gov/hmrc/exports/models/declaration/Transport.scala
```diff
@@ -42,4 +42,7 @@ case class Transport(
 
 object Transport {
   implicit val format: OFormat[Transport] = Json.format[Transport]
+
+  val optionNone = "option_none"
+  def typeApplicable(value: String): Boolean = value != optionNone
```
MaciejRewera: What do you think about moving this method to the class? You could then write `transport.isApplicable`

SteveSugden: That would be better.  Thanks

SteveSugden: Actually, it would have to be something like `def isMeansOfTransportOnDepartureApplicable: Boolean = meansOfTransportOnDepartureType.exists(_ != Transport.optionNone)` as there may be other values that are not applicable.

SteveSugden: Created method on Transport class

SteveSugden:
Not sure that I need to align with the frontend for the new option.  The BE doesn't seem to know anything about the other transport codes

PR #173, opened 2020-03-24
CEDS-2247 Introduce Country class for countries and migrate the database to the new model

PatrykRudnicki:


app/uk/gov/hmrc/exports/mongock/changesets/CacheChangeLog.scala
```diff
@@ -68,4 +67,80 @@ class CacheChangeLog {
...
+    val query: Document = new Document(queryParams)
+
+    val documents = db.getCollection(collection).find(new BasicDBObject(query)).toSeq
+
+    documents.foreach { document =>
+      val locations = document.get("locations", classOf[Map[String, String]])
+      val originationCountry: String = locations.get("originationCountry")
+
+      val queryIndexes: Map[String, String] =
+        ImmutableMap.of("id", document.get("id").asInstanceOf[String], "eori", document.get("eori").asInstanceOf[String])
```
SteveSugden: Why are we referencing "eori" here?

SteveSugden: Are id and eori keys to the document?

PatrykRudnicki: yeah, we treated this like identifier to easier find the documents.

Implemented with @PauloPortugal

SteveSugden: Thought so.  Maybe we could extract that bit out as we will need it whenever we update a declaration?

PatrykRudnicki: I tried this, but it's related with the information from specific document. I can extract this as private method if it will be more readable or helpful.

PR #182, opened 2020-04-01
CEDS-2250 DB migration

MaciejRewera:


app/uk/gov/hmrc/exports/mongock/changesets/CacheChangeLog.scala
```diff
@@ -151,9 +162,27 @@ class CacheChangeLog {
...
     }
     logger.info("Applying 'CEDS-2247 Change routing countries structure... Done.")
   }
+
+  @ChangeSet(order = "006", id = "CEDS-2250 Add one structure level to /transport/borderModeOfTransportCode", author = "Maciej Rewera")
+  def updateTransportBorderModeOfTransportCode(db: MongoDatabase): Unit = {
+
+    logger.info("Applying 'CEDS-2250 Add one structure level to /transport/borderModeOfTransportCode'...")
+
+    getDeclarationsCollection(db).updateMany(
```
PauloPortugal, line 175: Good! Whenever possible we need to be able to do these operations and delegate them to MongoDB
:+1:

app/uk/gov/hmrc/exports/mongock/changesets/CacheChangeLog.scala
```diff
@@ -120,27 +131,27 @@ class CacheChangeLog {
...
   @ChangeSet(order = "005", id = "CEDS-2247 Change routing countries structure", author = "Patryk Rudnicki")
   def changeRoutingCountriesStructure(db: MongoDatabase): Unit = {
+
     logger.info("Applying 'CEDS-2247 Change routing countries structure... ")
     val queryParams: ImmutableMap[String, AnyRef] =
       ImmutableMap.of("locations.routingCountries", ImmutableMap.of("$exists", true, "$ne", ""))
     val query: Document = new Document(queryParams)
 
-    val documents = db.getCollection(collection).find(new BasicDBObject(query)).toSeq
+    val documents = getDeclarationsCollection(db).find(new BasicDBObject(query))
```
PatrykRudnicki, line 147: I'm worried that this approach was wrong. This load all declarations to the memory and that was one of the problems on staging (around half of the million declarations).

Paulo's approach from JavaChangeLog is using mongo to perform updates, without loading the all declarations in that way.

MaciejRewera: Regarding loading eagerly/lazily this code behaves in exactly the same way as the Java code you mention. Except the difference that I am assigning the result to a value.
The `find` method returns `FindIterable` which (although not guaranteed in futurre versions) is taking documents lazily in batches (to limit amount of DB calls): https://stackoverflow.com/questions/39161042/is-finditerabledocument-load-all-documents

MaciejRewera:
The reason why I want to keep them is that mixing migrations for a single collection in several files is a bad ideae. Imagine scenario: we have 2 files, let's call them ChangeLog1 and ChangeLog2. We have 5 change sets (as we have now) in ChangeLog1. Then we add change set no. 6 in ChangeLog2 (and this is the only change set iin that file). So far so good. We can set the order in which files are executed and the change sets inside them as well.
Problem appears if we'd like to have another change set in ChangeLog1 (no. 7) and it is dependent on change set no. 6 (in ChangeLog2). I am not aware of any mechanism that would allow us to execute some of the change set from ChangeLog1, then from ChangeLog2 and back from ChangeLog1.

app/uk/gov/hmrc/exports/mongock/changesets/CacheChangeLog.scala
PatrykRudnicki: That's a good point, but Java methods right now filter more specifically what need to be updated.

I'll leave it to @PauloPortugal he spent more time on those migrations and knows more about it :)

PauloPortugal:
@MaciejRewera the scenario you present is possible, but it is very low risk. Normally `ChangeSet` operations tend to be independent of other `ChangeSet`s

I would then suggest that we create a Tech Debt ticket to migrate the Java MongoDriver approach to a ScalaMongoDriver, as there are operations and an approach that we would like to have it as an example on how to iterate through a collection of documents.

MaciejRewera:
Run all migrations using `mongo-scala-driver` and all passed successfully.

app/uk/gov/hmrc/exports/mongock/changesets/CacheChangeLog.scala
```diff
@@ -16,144 +16,35 @@
...
-    documents.foreach { document =>
-      val locations = document.get("locations", classOf[Map[String, String]])
-      val originationCountry: String = locations.get("originationCountry")
+    logger.info("Applying 'CEDS-2250 Add one structure level to /transport/borderModeOfTransportCode'...")
 
-      val queryIndexes: Map[String, String] =
-        ImmutableMap.of("id", document.get("id").asInstanceOf[String], "eori", document.get("eori").asInstanceOf[String])
+    getDeclarationsCollection(db).updateMany(
+      and(exists("transport.borderModeOfTransportCode"), not(Filters.eq("transport.borderModeOfTransportCode", ""))),
+      rename("transport.borderModeOfTransportCode", "temp")
```
PatrykRudnicki, line 39: I'm wondering why you're changing this to temp and after that once again name? Is not easier to just retrieve the value and update it?

With this approach you're retrieving the collection two times.

MaciejRewera: This is what you'd get:
`com.mongodb.MongoWriteException: The source and target field for $rename must not be on the same path: transport.borderModeOfTransportCode: "transport.borderModeOfTransportCode.code" `
So you have to create a temporary field.
And here, you are not retrieving collection at all. You only ask mongo to do update. It is true that it is twice, but still it is over 100 times faster than the solution with reading documents, updating them in the application and sending them for update to Mongo.

PR #183, opened 2020-04-02
CEDS-2250 Add tests for CacheChangeLog classes

MaciejRewera:
Except adding tests for migration functions, it also fixes ChangeSets no. 005 and 006.

In the first one the problem was that non-migrated documents were not recognized
correctly and therefore the migration was not applied to them.
The problem turned out to be a `type` parameter in the search query.

In the second one the problem was that the migration could potentially be executed
on documents that were already correctly migrated, adding another level of nesting 
with every call. Additional query parameter was added to prevent such situation.

### DIFF
--- a/app/uk/gov/hmrc/exports/mongock/changesets/CacheChangeLog.scala
+++ b/app/uk/gov/hmrc/exports/mongock/changesets/CacheChangeLog.scala
@@ -35,7 +35,11 @@ class CacheChangeLog {
     logger.info("Applying 'CEDS-2250 Add one structure level to /transport/borderModeOfTransportCode'...")
 
     getDeclarationsCollection(db).updateMany(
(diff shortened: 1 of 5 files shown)


app/uk/gov/hmrc/exports/mongock/changesets/CacheChangeLog.scala
```diff
@@ -35,7 +35,11 @@ class CacheChangeLog {
     logger.info("Applying 'CEDS-2250 Add one structure level to /transport/borderModeOfTransportCode'...")
 
     getDeclarationsCollection(db).updateMany(
-      and(exists("transport.borderModeOfTransportCode"), not(Filters.eq("transport.borderModeOfTransportCode", ""))),
+      and(
+        exists("transport.borderModeOfTransportCode"),
```
PauloPortugal, line 39: :+1:

PR #187, opened 2020-05-01
CEDS-2065 - add new 'representing another agent' field to model

SteveSugden:
Adding new field to RepresentativeDetails.  The field is not used in constructing the XML payload - it is purely there to drive frontend navigation and validation logic.


SteveSugden:
It should just work.  The new field is just for navigate.  The user would see the existing Eori and type answers in the summary screen and "blank" for the new question.  But they could still submit.  (Or change the answers)

PR #189, opened 2020-05-05
CEDS-2371 Add url to the notification pointers

PatrykRudnicki:


app/uk/gov/hmrc/exports/services/WCOPointerMappingService.scala
```diff
@@ -16,27 +16,47 @@
...
-      .map(_.split(","))
-      .map {
-        case Array(wcoPattern, exportsPattern) =>
-          PointerMapping(PointerPattern(wcoPattern.trim), PointerPattern(exportsPattern.trim))
-      }
-      .toSet
+
+import scala.io.Source
+
+object WCOPointerMappingService {
```
SteveSugden, line 25: Why the change away from class with injected file reader?

SteveSugden:
General question.  Why are we moving the addition of the url to the BE?  Are urls not a FE concern?

app/uk/gov/hmrc/exports/services/WCOPointerMappingService.scala
PatrykRudnicki: I had a problem with Injected one, Null Pointers.

Additionally I decided to use one approach, we used this approach in frontend application.

It contains static content, which is loaded during application startup, so I think object is okay

PatrykRudnicki:
@SteveSugden Urls are connected to the specific pointers. We have defined pointers in the backend application.

Of course I can create a new file and copy the pointers to the frontend application, but bot sure if it's worth.

SteveSugden:
It's probably because I don't understand the various sources of errors and code.
So.  In the BE we translate error codes into a pointer into out exports model.  E.g. 

42A.15A.T018 => declaration.borderTransport.meansOfTransportCrossingTheBorderType

Then in the front-end we use the messages file to translate pointers to field name (used on the error summary) E.g.

field.declaration.borderTransport.meansOfTransportCrossingTheBorderType => Active means of transport type

So I'm wondering why we can't translate pointer to url in the FE.

But in general your solution is fine

PR #192, opened 2020-05-13
CEDS-2406 Add IsExs model

PatrykRudnicki:


PatrykRudnicki:
Merging backend, optional fields, works with current frontend

PR #194, opened 2020-05-18
CEDS-2298 Display Time always on British time zone

PauloPortugal:
DMS sends the times in UTC time, but our BE microservice converts them
to `LocalDateTime` which strips them of the Zone information.

Converting the type to `Instant` as we did in the the Movements service
allows us to store the date from DMS as `UTC` or `Zulu time` and then on
the FE we just need to inform the view that we want to render that time
on British Time Zone `Europe/London`.

This ensures that we will see the time as `GMT/UTC` and `BST` depending
on the time of the year.

This holds true as long as DMS returns their dates in UTC.

This leaves us with a legacy time data in the `LocalDateTime` format.
I had two options:
 * Update the DB, by appending the `Z` (Zero Offset time AKA Zulu Time) to
   every timestamp. This is easily said than done, as the timestamps are
   within an array and the ineficciency of they MongoDB query update
   could prove an issue
 * Programatically deal with legacy and new data. This was the approach
   I chose.

Notification.scala
```
((__ \ "dateTimeIssued").read[Instant] or (__ \ "dateTimeIssued").read[Instant](readLocalDateTimeFromString))
```

Action.scala
```
((__ \ "requestTimestamp").read[Instant] or (__ \ "requestTimestamp").read[Instant](readLocalDateTimeFromString))
```

This allows the DAO layer to read two possible formats from the DB.

### DIFF
--- a/app/uk/gov/hmrc/exports/models/declaration/notifications/Notification.scala
+++ b/app/uk/gov/hmrc/exports/models/declaration/notifications/Notification.scala
@@ -16,20 +16,36 @@
 
 package uk.gov.hmrc.exports.models.declaration.notifications
 
-import java.time.LocalDateTime
+import java.time.{LocalDateTime, ZoneId, ZonedDateTime}
 
-import play.api.libs.json.{Json, OFormat}
+import play.api.libs.functional.syntax._
+import play.api.libs.json.{Json, Reads, _}
 import uk.gov.hmrc.exports.models.declaration.submissions.SubmissionStatus.SubmissionStatus
 
 case class Notification(
   actionId: String,
   mrn: String,
-  dateTimeIssued: LocalDateTime,
+  dateTimeIssued: ZonedDateTime,
   status: SubmissionStatus,
   errors: Seq[NotificationError],
   payload: String
 )
 
 object Notification {
-  implicit val format: OFormat[Notification] = Json.format[Notification]
+  implicit val readLocalDateTimeFromString: Reads[ZonedDateTime] = implicitly[Reads[LocalDateTime]]
+    .map(ZonedDateTime.of(_, ZoneId.of("UTC")))
+
+  implicit val writes: OWrites[Notification] = Json.writes[Notification]
+  implicit val reads: Reads[Notification] =
+    ((__ \ "actionId").read[String] and
+      (__ \ "mrn").read[String] and
+      ((__ \ "dateTimeIssued").read[ZonedDateTime] or (__ \ "dateTimeIssued").read[ZonedDateTime](readLocalDateTimeFromString)) and
+      (__ \ "status").read[SubmissionStatus] and
+      (__ \ "errors").read[Seq[NotificationError]] and
+      (__ \ "payload").read[String])(Notification.apply _)
+
+  implicit val notificationsReads: Reads[Seq[Notification]] = Reads.seq(reads)
+  implicit val notificationsWrites: Writes[Seq[Notification]] = Writes.seq(writes)
+
+  implicit val format: Format[Notification] = Format(reads, writes)
 }
--- a/app/uk/gov/hmrc/exports/models/declaration/submissions/Action.scala
+++ b/app/uk/gov/hmrc/exports/models/declaration/submissions/Action.scala
@@ -16,19 +16,20 @@
 
 package uk.gov.hmrc.exports.models.declaration.submissions
 
-import java.time.LocalDateTime
+import java.time.{LocalDateTime, ZoneId, ZonedDateTime}
 
-import play.api.libs.json.Json
+import play.api.libs.functional.syntax._
+import play.api.libs.json.{Json, Reads, _}
 
-/**
-  * Action class
-  *
-  * @param id unique id of the Action.  This is the converstationId from customs-notifications
-  * @param requestType the request type
-  * @param requestTimestamp timestamp (defaults to now)
-  */
-case class Action(id: String, requestType: RequestType, requestTimestamp: LocalDateTime = LocalDateTime.now())
+case class Action(id: String, requestType: RequestType, requestTimestamp: ZonedDateTime = ZonedDateTime.now(ZoneId.of("UTC")))
 
 object Action {
-  implicit val format = Json.format[Action]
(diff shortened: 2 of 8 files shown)


app/uk/gov/hmrc/exports/controllers/SubmissionController.scala
```diff
@@ -21,6 +21,7 @@ import play.api.mvc._
 import uk.gov.hmrc.exports.controllers.actions.Authenticator
 import uk.gov.hmrc.exports.controllers.response.ErrorResponse
 import uk.gov.hmrc.exports.controllers.util.HeaderValidator
+import uk.gov.hmrc.exports.models.declaration.submissions.Submission
```
PatrykRudnicki, line 24: Probably unnecessary import

app/uk/gov/hmrc/exports/models/declaration/notifications/Notification.scala
```diff
@@ -16,20 +16,36 @@
...
-  dateTimeIssued: LocalDateTime,
+  dateTimeIssued: Instant,
   status: SubmissionStatus,
   errors: Seq[NotificationError],
   payload: String
 )
 
 object Notification {
-  implicit val format: OFormat[Notification] = Json.format[Notification]
+  implicit val readLocalDateTimeFromString: Reads[Instant] = implicitly[Reads[LocalDateTime]]
```
PatrykRudnicki, line 35: I'm wondering if it's worth to extract this to some helper object. It's used in two places.

Additionally we can add documentation note with information why this etc. For future reference

app/uk/gov/hmrc/exports/models/declaration/notifications/Notification.scala
```diff
@@ -16,20 +16,36 @@
...
+    .map(_.atZone(ZoneOffset.UTC).toInstant)
+
+  implicit val writes: OWrites[Notification] = Json.writes[Notification]
+  implicit val reads: Reads[Notification] =
+    ((__ \ "actionId").read[String] and
+      (__ \ "mrn").read[String] and
+      ((__ \ "dateTimeIssued").read[Instant] or (__ \ "dateTimeIssued").read[Instant](readLocalDateTimeFromString)) and
```
MaciejRewera, line 42: 👍

app/uk/gov/hmrc/exports/services/NotificationService.scala
```diff
@@ -106,7 +106,7 @@ class NotificationService @Inject()(submissionRepository: SubmissionRepository,
         val mrn = (singleResponseXml \ "Declaration" \ "ID").text
         val formatter304 = DateTimeFormatter.ofPattern("yyyyMMddHHmmssX")
         val dateTimeIssued =
-          LocalDateTime.parse((singleResponseXml \ "IssueDateTime" \ "DateTimeString").text, formatter304)
+          LocalDateTime.parse((singleResponseXml \ "IssueDateTime" \ "DateTimeString").text, formatter304).atZone(ZoneOffset.UTC).toInstant
```
MaciejRewera, line 109: I wonder if `.atZone(ZoneOffset.UTC)` is necessary. We should have the timezone in the field, so parsing and `toInstant` should be enough.

app/uk/gov/hmrc/exports/models/declaration/notifications/Notification.scala
```diff
@@ -16,20 +16,36 @@
...
+  dateTimeIssued: Instant,
   status: SubmissionStatus,
   errors: Seq[NotificationError],
   payload: String
 )
 
 object Notification {
-  implicit val format: OFormat[Notification] = Json.format[Notification]
+  implicit val readLocalDateTimeFromString: Reads[Instant] = implicitly[Reads[LocalDateTime]]
+    .map(_.atZone(ZoneOffset.UTC).toInstant)
```
SteveSugden, line 36: Are we safe assuming the incoming dateTime is UTC?   If it has a timezone component would it not be better to read as Instant? (I.e. use the zone from the incoming data)

app/uk/gov/hmrc/exports/models/declaration/submissions/Action.scala
```diff
@@ -16,19 +16,20 @@
...
-  * @param requestType the request type
-  * @param requestTimestamp timestamp (defaults to now)
-  */
-case class Action(id: String, requestType: RequestType, requestTimestamp: LocalDateTime = LocalDateTime.now())
+case class Action(id: String, requestType: RequestType, requestTimestamp: Instant = Instant.now())
 
 object Action {
-  implicit val format = Json.format[Action]
+  implicit val readLocalDateTimeFromString: Reads[Instant] = implicitly[Reads[LocalDateTime]]
+    .map(_.atZone(ZoneOffset.UTC).toInstant)
```
SteveSugden, line 28: Similar comment

app/uk/gov/hmrc/exports/services/NotificationService.scala
SteveSugden: Inclined to agree

app/uk/gov/hmrc/exports/models/declaration/notifications/Notification.scala
PauloPortugal: This would be the values we have on our database, which are in 

> LocalDateTime

 format.
We are fine here

app/uk/gov/hmrc/exports/models/declaration/submissions/Action.scala
PauloPortugal: This would be the values we have on our database, which are in

`LocalDateTime`

format.
We are fine here

app/uk/gov/hmrc/exports/services/NotificationService.scala
PauloPortugal: This has been changed...

app/uk/gov/hmrc/exports/models/declaration/notifications/Notification.scala
PauloPortugal: The comments are on the commit messages :P, where they should live.
I actually prefer it in the code, as it is concise and easy to see.

PR #195, opened 2020-05-19
CEDS-2387 Add Id field to PackageInformation (with DB migration)

MaciejRewera:


app/uk/gov/hmrc/exports/models/declaration/ExportItem.scala
```diff
@@ -95,7 +96,7 @@ object IdentificationTypeCodes {
   }
 }
 
-case class PackageInformation(typesOfPackages: Option[String], numberOfPackages: Option[Int], shippingMarks: Option[String])
+case class PackageInformation(id: UUID, typesOfPackages: Option[String], numberOfPackages: Option[Int], shippingMarks: Option[String])
```
SteveSugden, line 99: Would it not be simpler to defined id as a String?   Would allow us to perhaps change the id generator for new package infos on the FE to something different in future (although I do note in your tests that UUIDs can be created from any old string)

MaciejRewera: In my opinion it is better to have it as `UUID` as it forces the client to use this type of ID.
I can see your point that having `String` is more flexible, but as you've said, you can create UUIDs from Strings with no problem.

app/uk/gov/hmrc/exports/models/generators/StringIdGenerator.scala
```diff
@@ -0,0 +1,28 @@
...
+
+import javax.inject.Singleton
+import org.apache.commons.lang3.RandomStringUtils
+
+@Singleton
+class StringIdGenerator extends IdGenerator[String] {
+  // http://commons.apache.org/proper/commons-lang/javadocs/api-3.6/org/apache/commons/lang3/RandomStringUtils.html
+  // no longer deprecated http://commons.apache.org/proper/commons-lang/javadocs/api-3.9/org/apache/commons/lang3/RandomStringUtils.html
+  // https://issues.apache.org/jira/browse/LANG-1346
+  def generateId(): String = RandomStringUtils.random(8, "0123456789abcdefg")
```
SteveSugden, line 27: Just curious why we use Apache Commons in the FE (given the comment).   
I've just used `(scala.util.)Random.alphanumeric.take(8).mkString.toLowerCase` in my PR

app/uk/gov/hmrc/exports/mongock/changesets/CacheChangeLog.scala
```diff
@@ -36,6 +38,10 @@ class CacheChangeLog {
   private val INDEX_ID = "id"
   private val INDEX_EORI = "eori"
 
+  private var idGenerator: IdGenerator[String] = new StringIdGenerator
```
PatrykRudnicki, line 41: Why `var`?

app/uk/gov/hmrc/exports/mongock/changesets/CacheChangeLog.scala
```diff
@@ -36,6 +38,10 @@ class CacheChangeLog {
...
+  private[changesets] def setIdGenerator(newIdGenerator: IdGenerator[String]): Unit =
+    this.idGenerator = newIdGenerator
```
PatrykRudnicki, line 43: This method is not used in this file. What is the reason of this?

test/it/uk/gov/hmrc/exports/mongock/changesets/CacheChangeLogSpec.scala
```diff
@@ -34,9 +37,13 @@ class CacheChangeLogSpec extends WordSpec with MustMatchers with GuiceOneServerP
 
   private val changeLog = new CacheChangeLog()
 
+  private val idGenerator = mock[IdGenerator[String]]
+  changeLog.setIdGenerator(idGenerator)
```
SteveSugden, line 41: It's used here @PatrykRudnicki - some sort of code injection. :-)

PatrykRudnicki: CacheChangeLog is a normal scala class. I think we can use it in a way like always `class CacheChangeLog @Inject()(idGenerator: IdGenerator)`.

Additionally creating a methods just for tests are probably not a good idea.

In general it's inside the cache changes/migrations, so it's not a big deal, so it's non blocking comment

MaciejRewera: You cannot use any Injection framework in ChangeLogs. Unfortunatelly, they are dependent on `Mongock` object and it is the expected behavior that ChangeLogs are indpendent from any framework.
I would agree with adding a method just for test, but I don't see much options here. Besides this class is not widely used, it's scope is limited so it should not do much harm to have it as it is.

PatrykRudnicki:
```
private val defaultGenerator: IdGenerator[String] = new StringIdGenerator
  private def idGenerator(generator: IdGenerator[String] = defaultGenerator): IdGenerator[String] = generator
```

In that case probably enough to use this approach

PatrykRudnicki: but this won't give a possibility to use it in tests. Okay, out of ideas right now

PR #213, opened 2020-06-26
CEDS-2557 migration spike

MaciejRewera:


app/uk/gov/hmrc/exports/migrations/ExportsMigrationTool.scala
```diff
@@ -0,0 +1,110 @@
...
+    val changeEntry = ChangeEntry(migrDefinition.migrationInformation)
+
+    try {
+      if (changeEntryRepository.isNewChange(changeEntry)) {
+        migrDefinition.migrationFunction(database)
+        changeEntryRepository.save(changeEntry)
+        logger.info(s"${changeEntry} applied")
+
+      } else if (isRunAlways(migrDefinition)) {
+        migrDefinition.migrationFunction(database)
```
SteveSugden, line 96: Minor point: Should this 'else if' block not be combined with the 'if'?   I.e. if new or run-always > run > update changeEntry db.   Otherwise (a) you have no record of when a run-always task last run and (b) if you switch run-always off it will run one more time as it will be "new"

MaciejRewera: I did not spend more time refactoring. This is almost 1-to-1 copy. Only changed that to scala from java.

SteveSugden: That's fine.   I thought the "run always" was your feature.  Didn't realise that Mongock did that.

PR #216, opened 2020-07-06
CEDS-2557 migration spike

MaciejRewera:


test/unit/uk/gov/hmrc/exports/migrations/ExportsMigrationToolSpec.scala
```diff
@@ -0,0 +1,248 @@
...
+      val testMigrationInformation = MigrationInformation(id = "TestId", order = 1, author = "TestAuthor")
+      val migrationDefinition = new MigrationDefinition {
+        override val migrationInformation: MigrationInformation = testMigrationInformation
+        override def migrationFunction(db: MongoDatabase): Unit = {}
+      }
+
+      "call ChangeEntryRepository findAll method" in {
+
+        when(migrationsRegistry.migrations).thenReturn(Seq(migrationDefinition))
+//        when(changeEntryRepository.isNewChange(any())).thenReturn(true)
```
SteveSugden, line 97: Delete commented out lines (+ other places)

test/unit/uk/gov/hmrc/exports/migrations/repositories/TestObjectsBuilder.scala
```diff
@@ -0,0 +1,34 @@
...
+
+import com.mongodb.client.MongoCursor
+import com.mongodb.{ServerAddress, ServerCursor}
+import org.bson.Document
+
+object TestObjectsBuilder {
+  def buildMongoCursor(elements: Seq[Document]): MongoCursor[Document] = new MongoCursor[Document] {
+    private val iterator = elements.iterator
+
+    override def close(): Unit = ???
```
SteveSugden, line 27: What's all this un-implemented stuff for?

MaciejRewera: I have not found a way to easily create `MongoCursor` to be able to return it from mocks. So I thought about mocking this as well, but in the end it turned out that in this case it is easier to create anonymous object with minimum implementation. Some of the methods are not implemented, because they are not used in tests. Besides, some of them are specific to DB operations, while I only need the iterating part of the class.

SteveSugden: Ok, seems reasonable.  Maybe add a comment to explain why these methods are "unfinished"?

MaciejRewera: Can do. But I'll wait until you finish the review, so that my pushes not interfere it.

SteveSugden: I've finished - go for it

PR #217, opened 2020-07-07
Revert "CEDS-2550 - new end-point for cloning an existing declaration…

SteveSugden:
… (#212)"

This reverts commit 9dbd7b9d


SteveSugden:
Correct.  Nothing was using this poor end-point.

PR #218, opened 2020-07-13
CEDS-2607 Libraries upgrade

PatrykRudnicki:


app/uk/gov/hmrc/exports/connectors/CustomsDeclarationsConnector.scala
```diff
@@ -104,16 +104,16 @@ class CustomsDeclarationsConnector @Inject()(appConfig: AppConfig, httpClient: H
         logger.debug(s"Response: ${response.status} => ${response.body}")
         getHttpResponseStatusType(response) match {
           case ApplicationErrorStatus =>
-            throw Upstream4xxResponse(
+            throw UpstreamErrorResponse(
```
PatrykRudnicki, line 107: `Upstream4xxResponse` and `Upstream5xxResponse` are deprecated.

This version with passing headers is deprecated too but silenced, so no warnings. Probably worth to check if we need those headers and if it's useful, if not we can remove this.

PR #220, opened 2020-07-21
Upgrade sbt, scala and play framework versions

PatrykRudnicki:


build.sbt
```diff
@@ -20,7 +20,8 @@ lazy val ComponentTest = config("component") extend Test
 
 def oneForkedJvmPerTest(tests: Seq[TestDefinition]): Seq[Group] = {
   tests map {
-    test => Group(test.name, Seq(test), SubProcess(ForkOptions(runJVMOptions = Seq("-Dtest.name=" + test.name))))
+    //test => Group(test.name, Seq(test), SubProcess(ForkOptions(runJVMOptions = Seq("-Dtest.name=" + test.name))))
+    test => Group(test.name, Seq(test), SubProcess(ForkOptions()))
```
PatrykRudnicki, line 24: Fork options has a different signature right now, I change this to default one. I'm not sure if we should use this method. The best option is probably to remove it, but wasn't sure if it's important

build.sbt
```diff
@@ -43,32 +44,32 @@ lazy val microservice = Project(appName, file("."))
...
     ),
     addTestReportOption(Test, "test-reports")
   )
   .settings(inConfig(IntegrationTest)(Defaults.itSettings): _*)
   .settings(
-    Keys.fork in IntegrationTest := false,
-    unmanagedSourceDirectories in IntegrationTest := Seq(
-      (baseDirectory in IntegrationTest).value / "test/it",
-      (baseDirectory in Test).value / "test/util"
+    IntegrationTest / Keys.fork := false,
```
PatrykRudnicki, line 55: Syntax change, now it's with `/` instead of `in`

project/AppDependencies.scala
```diff
@@ -7,9 +7,9 @@ object AppDependencies {
   private val testScope = "test, it, component"
 
   val compile = Seq(
-    "uk.gov.hmrc"                    %%  "simple-reactivemongo"  % "7.30.0-play-26",
+    "uk.gov.hmrc"                    %%  "simple-reactivemongo"  % "7.30.0-play-27",
```
PatrykRudnicki, line 10: Here I have a problem. Sbt for me is trying to download this library with scala version 2.11, which not exists, because play 2.7 require scala 2.12

project/plugins.sbt
```diff
@@ -12,14 +12,14 @@ addSbtPlugin("uk.gov.hmrc" % "sbt-distributables" % "2.0.0")
 
 addSbtPlugin("uk.gov.hmrc" % "sbt-artifactory" % "1.3.0")
 
-addSbtPlugin("com.typesafe.play" % "sbt-plugin" % "2.6.23")
+addSbtPlugin("com.typesafe.play" % "sbt-plugin" % "2.7.5")
```
PatrykRudnicki, line 15: Play framework version 2.7

PR #222, opened 2020-09-23
CEDS-2695 Remove userId from unit tests for auth migration

janeturner:
Remove deprecated SessionKeys userId attribute from unit tests


MaciejRewera (approved):
LGTM

PR #226, opened 2020-10-08
CEDS-2706 Add error pointers for Package Information

janeturner:


conf/code-lists/pointer-mapping.csv
```diff
@@ -68,7 +68,7 @@
 42A.67A.68A.$1.23A.70A.166,declaration.items.$1.procedureCodes.procedureCode.current,/customs-declare-exports/declaration/items/ITEM_ID/procedure-codes
 42A.67A.68A.$1.23A.70A.161,declaration.items.$1.procedureCodes.procedureCode.previous,/customs-declare-exports/declaration/items/ITEM_ID/procedure-codes
 42A.67A.68A.$1.23A.93A.$2.142,declaration.items.$1.packageInformation.$2.shippingMarks,
-42A.67A.68A.$1.23A.93A.$2.144,declaration.items.$1.packageInformation.$2.numberOfPackages,
+42A.67A.68A.$1.23A.93A.$2.144,declaration.items.$1.packageInformation.$2.numberOfPackages,/customs-declare-exports/declaration/items/ITEM_ID/package-information
```
MaciejRewera, line 71: This link may not be the best one. As you can see, you can have multiple packages added, but at the moment it is not possible to redirect directly to the page with particular package information. 
In this case I'd rather redirect to the page with the list of packages. This pattern is already in use and you can see it a few lines below.
In your current approach you'd force the user to add a new package every time.

Also, what about other package-information pointers? Should they not have url as well?

MaciejRewera (approved):

PR #231, opened 2020-11-03
CEDS-2755 Add a Pointer Based URL logic

PauloPortugal:
In some instances the WCO Declaration pointer will be returned but it
can map to two different pages on our user journey, which is depending on the
declaration type and the questions answered.

The approach taken is to provide a URL_PATH placeholder that will then
be replaced by a URL PATH resolver in the frontend.


conf/code-lists/pointer-mapping.csv
```diff
@@ -22,7 +22,7 @@
 42A.28A.62A.098,declaration.borderTransport.paymentMethod,/customs-declare-exports/declaration/transport-payment
 42A.28A.62A.$1.64,declaration.locations.destinationCountries.countriesOfRouting.$1,/customs-declare-exports/declaration/countries-summary/
 42A.40A.118,declaration.totalNumberOfItems.exchangeRate,/customs-declare-exports/declaration/total-numbers-of-items
-42A.57B.R123,declaration.declarantDetails.details.eori,/customs-declare-exports/declaration/person-presenting-goods
+42A.57B.R123,declaration.declarantDetails.details.eori,/customs-declare-exports/declaration/URL_PATH
```
glidester, line 25: Maybe surround placeholder with square or angled brackets to highlight it is not actually part of the url e.g [URL_PATH]?

MaciejRewera: If you do, could you change it for other placeholders too?

PauloPortugal: This has been addressed

glidester (approved):

PR #235, opened 2020-12-02
[CEDS-2800] Implement asynchronous notifications processing

glidester:
Before all Notification processing was a synchronous operation. In cases
where the service can not parse a Notification due to a change in its
structure, the notification data is lost as it is not inserted into
the MongoDB.

The implemented solution is:
1) The data parsed from a notification is now optional in the data
model.
2) All notifications payloads received are persisted to Mongo if they
are parsable or not.
3) A reattempt to parse 'routine' is run at app start up (after the
database migration jobs have finished) to query all unparsed
notifications and attempt to reprocess them.
4) A new database migration job created to migrate from the non-optional
notification parsed details model to the new optional structure.

### DIFF
--- a/app/uk/gov/hmrc/exports/migrations/changelogs/notification/MakeParsedDetailsOptional.scala
+++ b/app/uk/gov/hmrc/exports/migrations/changelogs/notification/MakeParsedDetailsOptional.scala
@@ -1,0 +1,102 @@
+/*
+ * Copyright 2020 HM Revenue & Customs
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package uk.gov.hmrc.exports.migrations.changelogs.notification
+
+import java.util
+
+import com.mongodb.client.MongoDatabase
+import org.bson.Document
+import org.mongodb.scala.bson.conversions.Bson
+import org.mongodb.scala.model.Filters.{and, exists, or, eq => feq}
+import org.mongodb.scala.model.UpdateOneModel
+import org.mongodb.scala.model.Updates.{combine, set, unset}
+import play.api.Logger
+import uk.gov.hmrc.exports.migrations.changelogs.{MigrationDefinition, MigrationInformation}
+
+import scala.collection.JavaConverters._
+
+class MakeParsedDetailsOptional extends MigrationDefinition {
+
+  private val logger = Logger(this.getClass)
+
+  private val INDEX_ID = "_id"
+  private val MRN = "mrn"
+  private val DATE_TIME_ISSUED = "dateTimeIssued"
+  private val STATUS = "status"
+  private val ERRORS = "errors"
+  private val DETAILS = "details"
+
+  private val collectionName = "notifications"
+
+  override val migrationInformation: MigrationInformation =
+    MigrationInformation(
+      id = "CEDS-2800 Make the parsed detail fields mrn, dateTimeIssued, status & error optional",
+      order = 1,
+      author = "Tim Wilkins",
+      runAlways = true
+    )
+
+  override def migrationFunction(db: MongoDatabase): Unit = {
+    logger.info(s"Applying '${migrationInformation.id}' db migration...")
+
+    val query = or(exists(MRN), exists(DATE_TIME_ISSUED), exists(STATUS), exists(ERRORS))
+    val queryBatchSize = 10
+    val updateBatchSize = 100
+
+    def createDetailsSubDoc(document: Document) = {
+      val mrn = document.get(MRN).asInstanceOf[String]
+      val dateTimeIssued = document.get(DATE_TIME_ISSUED).asInstanceOf[String]
+      val status = document.get(STATUS).asInstanceOf[String]
+      val errors = document.get(ERRORS, classOf[util.List[Document]])
+
+      new Document()
+        .append(MRN, mrn)
+        .append(DATE_TIME_ISSUED, dateTimeIssued)
+        .append(STATUS, status)
+        .append(ERRORS, errors)
+    }
+
+    getDocumentsToUpdate(db, query, queryBatchSize).map { document =>
+      val documentId = document.get(INDEX_ID)
+
+      val detailsSubDoc = createDetailsSubDoc(document)
+      logger.info(s"Creating new sub-document: $detailsSubDoc for documentId=$documentId")
+
+      val filter = and(feq(INDEX_ID, documentId))
+      val update = combine(unset(MRN), unset(DATE_TIME_ISSUED), unset(STATUS), unset(ERRORS), set(DETAILS, detailsSubDoc))
+      logger.info(s"[filter: $filter] [update: $update]")
+
+      new UpdateOneModel[Document](filter, update)
+    }.grouped(updateBatchSize).zipWithIndex.foreach {
+      case (requests, idx) =>
+        logger.info(s"Updating batch no. $idx...")
+
+        db.getCollection(collectionName).bulkWrite(seqAsJavaList(requests))
+        logger.info(s"Updated batch no. $idx")
+    }
+
+    logger.info(s"Applying '${migrationInformation.id}' db migration... Done.")
+  }
+
+  private def getDocumentsToUpdate(db: MongoDatabase, filter: Bson, queryBatchSize: Int) =
+    asScalaIterator(
+      db.getCollection(collectionName)
+        .find(filter)
+        .batchSize(queryBatchSize)
+        .iterator
+    )
+}
--- a/app/uk/gov/hmrc/exports/migrations/repositories/MongoRepository.scala
+++ b/app/uk/gov/hmrc/exports/migrations/repositories/MongoRepository.scala
@@ -53,14 +53,19 @@ abstract class MongoRepository private[migrations] (val mongoDatabase: MongoData
     }
 
   // Index is considered unique when:
-  // 1. Every uniqueField value is equal 1
-  // 2. "ns" field contains fullCollectionName
-  // 3. "unique" field is true
+  // 1. Name is "_id_" (these are implicitly unique https://docs.mongodb.com/manual/indexes/index.html#default-id-index)
+  // 2. Every uniqueField value is equal 1
+  // 3. "ns" field contains fullCollectionName
+  // 4. "unique" field is true
   private def isIndexUnique(index: Document): Boolean = {
     val key = index.get("key").asInstanceOf[Document]
+
+    if (index.getString("name").equals("_id_")) return true
+
     for (uniqueField <- uniqueFields) {
       if (key.getInteger(uniqueField, 0) != 1) return false
     }
+
     fullCollectionName == index.getString("ns") && index.getBoolean("unique", false)
   }
 
--- a/app/uk/gov/hmrc/exports/models/declaration/notifications/Notification.scala
+++ b/app/uk/gov/hmrc/exports/models/declaration/notifications/Notification.scala
@@ -22,30 +22,61 @@ import play.api.libs.functional.syntax._
 import play.api.libs.json.{Json, Reads, _}
 import uk.gov.hmrc.exports.models.declaration.submissions.SubmissionStatus.SubmissionStatus
 
-case class Notification(
-  actionId: String,
-  mrn: String,
-  dateTimeIssued: ZonedDateTime,
-  status: SubmissionStatus,
-  errors: Seq[NotificationError],
-  payload: String
-)
+case class NotificationDetails(mrn: String, dateTimeIssued: ZonedDateTime, status: SubmissionStatus, errors: Seq[NotificationError])
 
-object Notification {
+object NotificationDetails {
   implicit val readLocalDateTimeFromString: Reads[ZonedDateTime] = implicitly[Reads[LocalDateTime]]
     .map(ZonedDateTime.of(_, ZoneId.of("UTC")))
 
-  implicit val writes: OWrites[Notification] = Json.writes[Notification]
-  implicit val reads: Reads[Notification] =
-    ((__ \ "actionId").read[String] and
-      (__ \ "mrn").read[String] and
+  implicit val writes: OWrites[NotificationDetails] = Json.writes[NotificationDetails]
+  implicit val reads: Reads[NotificationDetails] =
+    ((__ \ "mrn").read[String] and
       ((__ \ "dateTimeIssued").read[ZonedDateTime] or (__ \ "dateTimeIssued").read[ZonedDateTime](readLocalDateTimeFromString)) and
       (__ \ "status").read[SubmissionStatus] and
-      (__ \ "errors").read[Seq[NotificationError]] and
-      (__ \ "payload").read[String])(Notification.apply _)
+      (__ \ "errors").read[Seq[NotificationError]])(NotificationDetails.apply _)
 
-  implicit val notificationsReads: Reads[Seq[Notification]] = Reads.seq(reads)
-  implicit val notificationsWrites: Writes[Seq[Notification]] = Writes.seq(writes)
+  implicit val format: Format[NotificationDetails] = Format(reads, writes)
+}
 
-  implicit val format: Format[Notification] = Format(reads, writes)
+case class Notification(actionId: String, payload: String, details: Option[NotificationDetails])
+
+object Notification {
+
+  object DbFormat {
+    implicit val writes: Writes[Notification] =
+      ((JsPath \ "actionId").write[String] and
+        (JsPath \ "payload").write[String] and
+        (JsPath \ "details").writeNullable[NotificationDetails](NotificationDetails.writes))(unlift(Notification.unapply))
+
+    implicit val reads: Reads[Notification] =
+      ((__ \ "actionId").read[String] and
+        (__ \ "payload").read[String] and
+        (__ \ "details").readNullable[NotificationDetails])(Notification.apply _)
+
+    implicit val notificationsReads: Reads[Seq[Notification]] = Reads.seq(reads)
+    implicit val notificationsWrites: Writes[Seq[Notification]] = Writes.seq(writes)
+
+    implicit val format: Format[Notification] = Format(reads, writes)
+  }
+
+  object FrontendFormat {
+    implicit val writes: Writes[Notification] = new Writes[Notification] {
+
+      def writes(notification: Notification) =
+        notification.details.map { details =>
+          Json.obj(
+            "actionId" -> notification.actionId,
+            "mrn" -> details.mrn,
+            "dateTimeIssued" -> details.dateTimeIssued,
+            "status" -> details.status,
+            "errors" -> details.errors,
+            "payload" -> notification.payload
+          )
+        }.getOrElse {
+          Json.obj("actionId" -> notification.actionId, "payload" -> notification.payload)
+        }
+    }
+
+    implicit val notificationsWrites: Writes[Seq[Notification]] = Writes.seq(writes)
+  }
 }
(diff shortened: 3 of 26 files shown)


PauloPortugal:
Great commit message! :100:

app/uk/gov/hmrc/exports/services/NotificationService.scala
```diff
@@ -39,15 +39,23 @@ class NotificationService @Inject()(submissionRepository: SubmissionRepository,
...
+          })
+      )
+
+      notification.copy(details = updatedDetails)
     }
+
+    val conversationIds = submission.actions.map(_.id)
+    notificationRepository
+      .findNotificationsByActionIds(conversationIds)
+      .map(_.map(updateErrorUrls(_)))
```
PauloPortugal, line 58: should we use `flatMap` here?

precheck.sh
```diff
@@ -1,3 +1,3 @@
 #!/usr/bin/env bash
 
-sbt clean coverage test it:test scalafmt::test test:scalafmt::test coverageReport
+sbt clean coverage test it:test scalafmt::test test:scalafmt::test scalastyle coverageReport
```
PauloPortugal, line 3: :thinking: Probably we could move `scalafmt::test test:scalafmt::test` as the first step, so that it fails fast.

app/uk/gov/hmrc/exports/migrations/changelogs/notification/MakeParsedDetailsOptional.scala
```diff
@@ -0,0 +1,102 @@
...
+    }
+
+    getDocumentsToUpdate(db, query, queryBatchSize).map { document =>
+      val documentId = document.get(INDEX_ID)
+
+      val detailsSubDoc = createDetailsSubDoc(document)
+      logger.info(s"Creating new sub-document: $detailsSubDoc for documentId=$documentId")
+
+      val filter = and(feq(INDEX_ID, documentId))
+      val update = combine(unset(MRN), unset(DATE_TIME_ISSUED), unset(STATUS), unset(ERRORS), set(DETAILS, detailsSubDoc))
```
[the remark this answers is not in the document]
glidester: Because they are now held in the DETAILS sub document

app/uk/gov/hmrc/exports/migrations/repositories/MongoRepository.scala
```diff
@@ -53,14 +53,19 @@ abstract class MongoRepository private[migrations] (val mongoDatabase: MongoData
...
-  // 2. "ns" field contains fullCollectionName
-  // 3. "unique" field is true
+  // 1. Name is "_id_" (these are implicitly unique https://docs.mongodb.com/manual/indexes/index.html#default-id-index)
+  // 2. Every uniqueField value is equal 1
+  // 3. "ns" field contains fullCollectionName
+  // 4. "unique" field is true
   private def isIndexUnique(index: Document): Boolean = {
     val key = index.get("key").asInstanceOf[Document]
+
+    if (index.getString("name").equals("_id_")) return true
```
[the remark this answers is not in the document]
glidester: Its not adding an index. This is my fix to the existing migration code that was breaking because this function was reporting the automatic mongo '_id' index as not being unique. That resulted in the migration code trying to drop the '_id' index and that isn't allowed (so the whole thing failed).

app/uk/gov/hmrc/exports/services/NotificationService.scala
glidester: First map is on a Future, second one is on a Seq.

precheck.sh
glidester: Perhaps, but if the tests fail and you have to change your code and you want the code to function correctly before you care about its style! :) Number of times my pipeline failed because of some silly scalastyle contravention I put this here so I wouldn't forget in future! :D

app/uk/gov/hmrc/exports/migrations/MigrationRunner.scala
```diff
@@ -51,12 +54,18 @@ class MigrationRunner @Inject()(
       logger.info("Exports Migration feature disabled. Starting migration with Mongock")
       migrateWithMongock()
     }
+
+    logger.info("Starting NotificationService.reattemptParsingUnparsedNotifications()...")
+    notificationService.reattemptParsingUnparsedNotifications().map { _ =>
+      logger.info("Finished NotificationService.reattemptParsingUnparsedNotifications()")
+    }
```
MaciejRewera, line 61: I'd rather extract this piece of code. It is not actually part of the migration. I guess it was the easiest way to ensure it is run after the migration, but logically it is not related with it.
I'd suggest to create a class with startup operations that would handle running both migration and parsing recovery. You'd need to change `ExportsMigrationModule` as well forthis purpose.

app/uk/gov/hmrc/exports/migrations/changelogs/notification/MakeParsedDetailsOptional.scala
```diff
@@ -0,0 +1,102 @@
...
+  private val STATUS = "status"
+  private val ERRORS = "errors"
+  private val DETAILS = "details"
+
+  private val collectionName = "notifications"
+
+  override val migrationInformation: MigrationInformation =
+    MigrationInformation(
+      id = "CEDS-2800 Make the parsed detail fields mrn, dateTimeIssued, status & error optional",
+      order = 9,
```
MaciejRewera, line 48: I'm wondering if it couldn't be iterated from 1. Mongock history is separate, and it seems no migration has been executed with new tool yet.

app/uk/gov/hmrc/exports/models/declaration/notifications/Notification.scala
```diff
@@ -22,30 +22,55 @@ import play.api.libs.functional.syntax._
...
+    implicit val writes: Writes[Notification] = new Writes[Notification] {
+      def writes(notification: Notification) = Json.obj(
+        "actionId" -> notification.actionId,
+        "mrn" -> notification.details.get.mrn,
+        "dateTimeIssued" -> notification.details.get.dateTimeIssued,
+        "status" -> notification.details.get.status,
+        "errors" -> notification.details.get.errors,
+        "payload" -> notification.payload
+      )
+    }
```
MaciejRewera, line 72: I assume there are no `reads` because we never read Notifications from the frontend?

app/uk/gov/hmrc/exports/models/declaration/notifications/Notification.scala
```diff
@@ -22,30 +22,55 @@ import play.api.libs.functional.syntax._
...
+
+    implicit val notificationsReads: Reads[Seq[Notification]] = Reads.seq(reads)
+    implicit val notificationsWrites: Writes[Seq[Notification]] = Writes.seq(writes)
+
+    implicit val format: Format[Notification] = Format(reads, writes)
+  }
 
-  implicit val notificationsReads: Reads[Seq[Notification]] = Reads.seq(reads)
-  implicit val notificationsWrites: Writes[Seq[Notification]] = Writes.seq(writes)
+  object frontendFormat {
```
MaciejRewera, line 62: Shouldn't be `FrontendFormat` ? ;-)

test/it/uk/gov/hmrc/exports/migrations/changelogs/notification/MakeParsedDetailsOptionalSpec.scala
```diff
@@ -0,0 +1,125 @@
...
+
+    "correctly migrate records from previous format" in {
+      runTest(testDataBeforeChangeSet_1, testDataAfterChangeSet_1)(changeLog.migrationFunction)
+    }
+
+    "not change records already migrated" in {
+      runTest(testDataAfterChangeSet_1, testDataAfterChangeSet_1)(changeLog.migrationFunction)
+    }
+
+    "not change records that could not be parsed" in {
```
MaciejRewera, line 58: For some reason this test name mislead me. At first glance I thought it is stating that we are not going to migrate what we cannot parse. 
It is not super-crucial, but what do you think about "not change records that were not yet parsed" ?

app/uk/gov/hmrc/exports/migrations/changelogs/notification/MakeParsedDetailsOptional.scala
glidester: Ok set to '1'

app/uk/gov/hmrc/exports/models/declaration/notifications/Notification.scala
glidester: Correct, it is just for us to send notifications in the old structured style to the front-end

app/uk/gov/hmrc/exports/models/declaration/notifications/Notification.scala
glidester: :) Yes. Done

test/it/uk/gov/hmrc/exports/repositories/NotificationRepositorySpec.scala
```diff
@@ -165,10 +175,87 @@ class NotificationRepositorySpec
...
+        val foundNotifications = repo.findUnparsedNotifications().futureValue
+
+        foundNotifications.length must equal(1)
+        foundNotifications must contain(notificationUnparsable)
+      }
+    }
+
+    "there are many unparsed Notification" should {
+      "return those Notifications" in {
+        val anotherUnparsableNotification = notificationUnparsable.copy(actionId = "c1c09f1b-7c94-4e90-b754-7c5c71c55e44")
```
MaciejRewera, line 209: Nit: In most our repositories there is an object `CommonTestData`. It contains examples on various identifiers, like DUCR, MUCR, MRN, ConversationId. You can use it so that you don't need to hard code them in tests.

test/it/uk/gov/hmrc/exports/migrations/changelogs/notification/MakeParsedDetailsOptionalSpec.scala
glidester: Ok, Renamed

app/uk/gov/hmrc/exports/migrations/repositories/MongoRepository.scala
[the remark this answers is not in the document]
PauloPortugal: > because this function was reporting the automatic mongo '_id' index as not being unique. 
mongo _id is always unique. How did we end up with _id duplicates?

test/it/uk/gov/hmrc/exports/repositories/NotificationRepositorySpec.scala
glidester: Ok, just wanted something that wasn't already used. I've modified it to use a pre-defined value from ExportsTestData now.

test/unit/uk/gov/hmrc/exports/services/NotificationServiceSpec.scala
```diff
@@ -175,12 +182,59 @@ class NotificationServiceSpec extends WordSpec with MockitoSugar with ScalaFutur
...
       }
     }
 
+    "NotificationService on reattemptParsingUnparsedNotifications" should {
+      "do nothing if no unparsed notifications exist" in new SaveHappyPathTest {
+        when(notificationRepositoryMock.findUnparsedNotifications())
+          .thenReturn(Future.successful(Seq.empty))
+
+        notificationService.reattemptParsingUnparsedNotifications().futureValue
+      }
```
MaciejRewera, line 199: If I see correctly, this test does not contain any assertion.

app/uk/gov/hmrc/exports/services/NotificationService.scala
```diff
@@ -84,21 +92,26 @@ class NotificationService @Inject()(submissionRepository: SubmissionRepository,
...
+          Future.successful(Left(exc.getMessage))
       }
-    } catch {
-      case exc: Throwable =>
-        logger.error(exc.getMessage)
-        Future.successful(Left(exc.getMessage))
-    }
+    }.getOrElse(Future.successful(Right((): Unit)))
 
   def buildNotificationsFromRequest(notificationApiRequestHeaders: NotificationApiRequestHeaders, notificationXml: NodeSeq): Seq[Notification] =
```
MaciejRewera, line 111: I'm wondering if this responsibility should not be extracted into separate builder or converter class.

test/unit/uk/gov/hmrc/exports/services/NotificationServiceSpec.scala
```diff
@@ -300,6 +354,28 @@ class NotificationServiceSpec extends WordSpec with MockitoSugar with ScalaFutur
...
+      output.size mustBe 2
+      output.foreach(_.details.isDefined mustBe true)
+    }
+
+    "return a single notification when the whole xml could not be successfully parsed" in new Test {
+      val output = notificationService.buildNotificationsFromRequest(headers, exampleNotificationWithUnparsableXML(mrn))
+      output.size mustBe 1
+      output.head.details.isDefined mustBe false
+    }
+  }
```
MaciejRewera, line 378: 👍 Thank you adding these.

app/uk/gov/hmrc/exports/services/NotificationService.scala
```diff
@@ -178,4 +194,24 @@ class NotificationService @Inject()(submissionRepository: SubmissionRepository,
...
+    notificationRepository
+      .findUnparsedNotifications()
+      .map(_.foreach { unparsedNotification =>
+        try {
+          val notifications = parseNotificationsPayload(unparsedNotification.actionId, XML.loadString(unparsedNotification.payload))
+
+          //successfully parsed the previously unparsable notification
+          notifications.head.details.map { _ =>
+            notificationRepository.removeNotificationsForActionId(unparsedNotification.actionId)
+            saveAll(notifications)
```
MaciejRewera, line 208: What do you think about swaping the order for these operations? To save first and then to remove the old one? Or even better, to execute them both as a single bulk operation?
The reason for this is that in the current order we remove the old Notification before making sure new ones are persisted.

MaciejRewera (commented):
Really good job. I like attention to details and fixes for parts not necessarily related to the ticket itself.

Left some comments as they may be highlighting some potential issues.

test/unit/uk/gov/hmrc/exports/services/NotificationServiceSpec.scala
glidester: Lol, well it testing doing nothing! Fixed :)

app/uk/gov/hmrc/exports/migrations/repositories/MongoRepository.scala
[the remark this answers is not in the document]
glidester: Its not a duplicate, it is the original _id index that this method was mistakenly saying is not unique.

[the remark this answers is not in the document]
PauloPortugal: So this is just being used to manage the `Changelog` collection?
Sorry... this is really confusing.

PauloPortugal:
I will leave this between you and @MaciejRewera. 
Looks good, although the migrations infrastructure code became quite complex.

app/uk/gov/hmrc/exports/services/NotificationService.scala
glidester: Yes this is a good spot! I've altered my code to save first and delete after, this way it is only vulnerable to creating duplicate notifications on system crashes. 

I've been looking at the current code and it seems there is currently an opportunity for notifications to be duplicated in the event of a system crash and re-delivery. The batch idea is a good one to prevent this but would mean reworking the existing notification controller logic also. Maybe a tech debt ticket in the making?

app/uk/gov/hmrc/exports/models/declaration/notifications/Notification.scala
```diff
@@ -22,30 +22,55 @@ import play.api.libs.functional.syntax._
...
-  implicit val notificationsReads: Reads[Seq[Notification]] = Reads.seq(reads)
-  implicit val notificationsWrites: Writes[Seq[Notification]] = Writes.seq(writes)
+  object FrontendFormat {
+    implicit val writes: Writes[Notification] = new Writes[Notification] {
+      def writes(notification: Notification) = Json.obj(
+        "actionId" -> notification.actionId,
+        "mrn" -> notification.details.get.mrn,
+        "dateTimeIssued" -> notification.details.get.dateTimeIssued,
+        "status" -> notification.details.get.status,
+        "errors" -> notification.details.get.errors,
```
MaciejRewera, line 69: @glidester Just realized that you're using `get` on `Option` here. I'm concerned a bit about it. I know there is a filter in controller not to send Notifications without `details`, but it is still error prone.

app/uk/gov/hmrc/exports/routines/RoutineRunner.scala
```diff
@@ -0,0 +1,42 @@
...
+
+import akka.actor.{ActorSystem, Cancellable}
+import javax.inject.Inject
+import play.api.inject.ApplicationLifecycle
+import uk.gov.hmrc.exports.migrations.MigrationRoutine
+
+import scala.concurrent.Future
+import scala.concurrent.duration._
+
+class RoutineRunner @Inject()(
```
MaciejRewera, line 27: 👍

app/uk/gov/hmrc/exports/routines/RoutineRunnerModule.scala
```diff
@@ -14,8 +14,8 @@
  * limitations under the License.
  */
 
-package uk.gov.hmrc.exports.migrations
+package uk.gov.hmrc.exports.routines
 
 import play.api.inject._
 
-class ExportsMigrationModule extends SimpleModule(bind[MigrationRunner].toSelf.eagerly())
+class RoutineRunnerModule extends SimpleModule(bind[RoutineRunner].toSelf.eagerly())
```
MaciejRewera, line 21: I can't find the new Module being registered in `application.conf`

glidester: Line 55

app/uk/gov/hmrc/exports/services/NotificationService.scala
```diff
@@ -178,4 +194,24 @@ class NotificationService @Inject()(submissionRepository: SubmissionRepository,
...
+  def reattemptParsingUnparsedNotifications(): Future[Unit] =
+    notificationRepository
+      .findUnparsedNotifications()
+      .map(_.foreach { unparsedNotification =>
+        try {
+          val notifications = parseNotificationsPayload(unparsedNotification.actionId, XML.loadString(unparsedNotification.payload))
+
+          //successfully parsed the previously unparsable notification
+          notifications.head.details.map { _ =>
+            saveAll(notifications)
```
MaciejRewera, line 207: Can it possibly happen that there are no Notifications in the list? Calling `head` would result in exception then. Maybe it would be safer to use `headOption` and then flatten the `Option`s ?

test/unit/uk/gov/hmrc/exports/services/NotificationServiceSpec.scala
```diff
@@ -175,12 +182,62 @@ class NotificationServiceSpec extends WordSpec with MockitoSugar with ScalaFutur
...
+          payload = NotificationTestData.exampleNotificationWithMultipleResponsesXML(mrn, dateTimeIssuedString).toString()
+        )
+
+        when(notificationRepositoryMock.findUnparsedNotifications())
+          .thenReturn(Future.successful(Seq(notificationNowParsable)))
+
+        notificationService.reattemptParsingUnparsedNotifications().futureValue
+
+        verify(notificationRepositoryMock, times(1)).removeUnparsedNotificationsForActionId(meq(notificationUnparsable.actionId))
+        verify(notificationRepositoryMock, times(2)).save(any())
```
MaciejRewera, line 232: Nit: I like to use `org.mockito.InOrder` in case I want to ensure the order of calls (like here you want to save before removing).
Just saying ;-)

MaciejRewera (commented):
Looks good, although I am not sure about this `notifications.head` part.

app/uk/gov/hmrc/exports/models/declaration/notifications/Notification.scala
glidester: Fixed

app/uk/gov/hmrc/exports/services/NotificationService.scala
glidester: Fixed

test/unit/uk/gov/hmrc/exports/services/NotificationServiceSpec.scala
glidester: Added

MaciejRewera (approved):
LGTM

PR #236, opened 2020-12-08
[CEDS-2800] Fix Notification index detailsMissingIdx filter

glidester:
Index was incorrectly set to just contain notifications where
the details sub-document is present.

Now it is correctly set to only contain notifications where
the details sub-document is not present.


MaciejRewera (approved):
LGTM

PR #241, opened 2020-12-14
[CEDS-2800] Drop misconfigured index

glidester:
Original definition of index 'detailsMissingIdx' used the
wrong filter (where doc exists rather than where doc does
not exist!)

Dropping index 'detailsMissingIdx' and re-creating it with
new fliter as 'detailsDocMissingIdx'.


MaciejRewera (approved):
Great naming!

PR #248, opened 2021-02-03
CEDS-3010 notifications parsing exception handling

MaciejRewera:
This change ensures that any Notification parsing
exceptions are caught and hot propagated to the 
caller.
Calling service may still receive 5xx response from
Exports, but only due to other issues than parsing.
The process is now also more aligned with the 
Movements service.

### DIFF
--- a/app/uk/gov/hmrc/exports/services/NotificationService.scala
+++ b/app/uk/gov/hmrc/exports/services/NotificationService.scala
@@ -16,24 +16,22 @@
 
 package uk.gov.hmrc.exports.services
 
-import java.time.format.DateTimeFormatter
-import java.time.{LocalDateTime, ZoneId, ZonedDateTime}
-
 import javax.inject.{Inject, Singleton}
 import play.api.Logger
 import reactivemongo.core.errors.DatabaseException
-import uk.gov.hmrc.exports.models.declaration.notifications.{Notification, NotificationDetails, NotificationError}
-import uk.gov.hmrc.exports.models.declaration.submissions.{Submission, SubmissionStatus}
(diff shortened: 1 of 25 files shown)


test/util/testdata/notifications/ExampleXmlAndDomainModelPair.scala
```diff
@@ -0,0 +1,24 @@
...
+ */
+
+package testdata.notifications
+
+import scala.xml.Elem
+
+trait ExampleXmlAndDomainModelPair[T] {
+  val asXml: Elem
+  val asDomainModel: T
+}
```
glidester, line 24: :+1:

app/uk/gov/hmrc/exports/services/NotificationService.scala
```diff
@@ -16,24 +16,22 @@
...
-  implicit executionContext: ExecutionContext
-) {
+class NotificationService @Inject()(
+  submissionRepository: SubmissionRepository,
+  notificationRepository: NotificationRepository,
+  notificationFactory: NotificationFactory
+)(implicit executionContext: ExecutionContext) {
 
   private val logger = Logger(this.getClass)
   private val databaseDuplicateKeyErrorCode = 11000
```
udinobi, line 37: I understand this is not related to this PR, but why do we see this value here? A 'Service' should know nothing of a such low-level db-related code.

glidester (dismissed):
LGTM

app/uk/gov/hmrc/exports/services/NotificationService.scala
MaciejRewera: Don't remember exactly why it is service handling the exception and not the repository class. It might be due to reusing one of already existing methods in repository that is more generic and should not handle the exception in all scenarios.

udinobi (approved):

PR #252, opened 2021-02-10
CEDS-2976 Add a connector to send emails via "email" service

udinobi:


app/uk/gov/hmrc/exports/models/emails/SendEmailRequest.scala
```diff
@@ -0,0 +1,25 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package uk.gov.hmrc.exports.models.emails
+
+import play.api.libs.json.Json
+
+case class SendEmailRequest(to: List[String], templateId: String, parameters: Map[String, String], eventUrl: Option[String] = None)
```
glidester, line 21: Might be a good idea to create an enum type for valid `templateId' values and `parameters` keys. Minimise the chance of passing garbage values.

udinobi: :+1:

test/it/uk/gov/hmrc/exports/connector/EmailConnectorSpec.scala
```diff
@@ -0,0 +1,77 @@
...
+import uk.gov.hmrc.exports.models.emails.EmailParameter.MRN
+import uk.gov.hmrc.exports.models.emails.TemplateId.DMSDOC_NOTIFICATION
+import uk.gov.hmrc.exports.models.emails.{EmailParameters, SendEmailError, SendEmailRequest}
+import uk.gov.hmrc.http.HttpClient
+
+class EmailConnectorSpec extends IntegrationTestSpec {
+
+  implicit val appConfig: AppConfig = inject[AppConfig]
+
+  val connector = new EmailConnector(inject[HttpClient])(appConfig, global)
```
MaciejRewera, line 34: You can simply `inject[EmailConnector]`

app/uk/gov/hmrc/exports/models/emails/SendEmailError.scala
```diff
@@ -0,0 +1,30 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package uk.gov.hmrc.exports.models.emails
+
+import play.api.libs.json.Json
+import uk.gov.hmrc.http.{HttpResponse, UpstreamErrorResponse}
+
+case class SendEmailError(status: Int, message: String)
```
MaciejRewera, line 22: I am not I understand what can be gained by introducing this class? `EmailConnector` could return `HttpResponse` and this way provide some information about successful case to the client as well.

test/unit/uk/gov/hmrc/exports/config/AppConfigSpec.scala
```diff
@@ -46,6 +46,8 @@ class AppConfigSpec extends UnitSpec {
       |microservice.services.customs-data-store.host=localhost
       |microservice.services.customs-data-store.port=6790
       |microservice.services.customs-data-store.verified-email-path=/customs-data-store/eori/EORI/verified-email
+      |microservice.services.hmrc-email.host=localhost
+      |microservice.services.hmrc-email.port=8300
```
MaciejRewera, line 50: Could you add tests for these as well? Why would this coonfig be added here?

app/uk/gov/hmrc/exports/models/emails/SendEmailError.scala
udinobi: Sorry, do not agree. Returning HttpResponse would move the logic to verify the response, which could be a success or a failure, to the client. What's wrong with doing that within the connector?

test/it/uk/gov/hmrc/exports/connector/EmailConnectorSpec.scala
udinobi: I find this more explicit. I do not have to consider if AppConfig is a singleton or not.

test/unit/uk/gov/hmrc/exports/config/AppConfigSpec.scala
udinobi: The tests in this spec class seem to me completely artificial. I do not understand what they are for. I tried to set a fake/wrong value in application.conf, which is where AppConfig takes the values from, and no one of the tests in this class raised any error. Can u please explain me what we are trying to verify here? Sorry, sometimes I'm a bit dumb :smiley:

app/uk/gov/hmrc/exports/models/emails/SendEmailError.scala
glidester: I agree, we should not leak HttpResponse outside of a connector unless there is a compelling reason to.

test/unit/uk/gov/hmrc/exports/config/AppConfigSpec.scala
```diff
@@ -16,125 +16,86 @@
...
-      |microservice.services.customs-declarations-information.host=localhostd
-      |microservice.services.customs-declarations-information.port=9834
-      |microservice.services.customs-declarations-information.api-version=1.0
-      |microservice.services.customs-declarations-information.bearer-token=Bearer cdi-bearer-token
-      |microservice.services.customs-declarations-information.client-id=cdi-client-id
-      |microservice.services.customs-declarations-information.fetch-mrn-status=/mrn/ID/status
-      |microservice.services.customs-data-store.host=localhost
-      |microservice.services.customs-data-store.port=6790
-      |microservice.services.customs-data-store.verified-email-path=/customs-data-store/eori/EORI/verified-email
-    """.stripMargin)
```
glidester, line 49: It looks like you've changed the purpose of this test class to now test the actual values in application.conf (rather than as before it was just testing the functionality of AppConfig in retrieving values or throwing exceptions). This means changing a value in application.conf will now also need to be reflected in this test also.

I agree though in that I don't see the point in these tests as they are mostly exercising the Play Bootstrap library rather than our own code. :thinking:

udinobi: > This means changing a value in application.conf will now also need to be reflected in this test also.
I think this is a good thing :smiley: and a more effective way to test the values loaded by AppConfig are there and the expected ones.

glidester (approved):

PR #253, opened 2021-02-11
CEDS-3036 asynchronous notification processing

MaciejRewera:
Notification parsing and then saving is now done separately from
responding to the request but it is executed in the same
ExecutionContext. Should it prove inefficient, new ExecutionContext can
be implemented at a later date.

Parsing and saving logic has been extracted from NotificationService as
this class was carrying too many responsibilities. It is now much easier
to test as well.

Being aware of incoming changes and new processes to be introduced on
the receipt of a new Notification, trait NotificationReceiptAction has
been implemented. This allows to have common interface for all these
processes.

### DIFF
--- a/app/uk/gov/hmrc/exports/connectors/CustomsDataStoreConnector.scala
+++ b/app/uk/gov/hmrc/exports/connectors/CustomsDataStoreConnector.scala
@@ -17,21 +17,20 @@
 package uk.gov.hmrc.exports.connectors
 
 import javax.inject.Inject
-
-import scala.concurrent.{ExecutionContext, Future}
-
+import play.api.http.Status.NOT_FOUND
 import uk.gov.hmrc.exports.config.AppConfig
 import uk.gov.hmrc.exports.models.VerifiedEmailAddress
-import play.api.http.Status.NOT_FOUND
-import uk.gov.hmrc.http.{HeaderCarrier, HttpClient, Upstream4xxResponse}
 import uk.gov.hmrc.http.HttpReads.Implicits._
+import uk.gov.hmrc.http.{HeaderCarrier, HttpClient, Upstream4xxResponse}
+
+import scala.concurrent.{ExecutionContext, Future}
 
 class CustomsDataStoreConnector @Inject()(http: HttpClient)(implicit appConfig: AppConfig, ec: ExecutionContext) {
 
   import CustomsDataStoreConnector._
 
   def getEmailAddress(eori: String)(implicit hc: HeaderCarrier): Future[Option[VerifiedEmailAddress]] =
-    http.GET[VerifiedEmailAddress](verifiedEmailUrl(eori)).map(Some(_)).recover {
+    http.GET[VerifiedEmailAddress](verifiedEmailUrl(eori)).map(Option(_)).recover {
       case Upstream4xxResponse(_, NOT_FOUND, _, _) => None
     }
 }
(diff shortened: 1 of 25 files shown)


app/uk/gov/hmrc/exports/services/notifications/NotificationService.scala
```diff
@@ -67,43 +68,23 @@ class NotificationService @Inject()(
...
+    }
+  }
 
-  private def saveSingleNotification(notification: Notification): Future[Unit] =
-    for {
-      _ <- notificationRepository.insert(notification)
-      _ <- updateRelatedSubmission(notification)
-    } yield (())
+  private def performActionsOnReceipt(notification: Notification): Cancellable = {
+    val notificationActions: Seq[NotificationReceiptAction] = Seq(parseAndSaveAction)
```
glidester, line 80: :+1:

app/uk/gov/hmrc/exports/connectors/CustomsDataStoreConnector.scala
```diff
@@ -17,21 +17,20 @@
...
+
+import scala.concurrent.{ExecutionContext, Future}
 
 class CustomsDataStoreConnector @Inject()(http: HttpClient)(implicit appConfig: AppConfig, ec: ExecutionContext) {
 
   import CustomsDataStoreConnector._
 
   def getEmailAddress(eori: String)(implicit hc: HeaderCarrier): Future[Option[VerifiedEmailAddress]] =
-    http.GET[VerifiedEmailAddress](verifiedEmailUrl(eori)).map(Some(_)).recover {
+    http.GET[VerifiedEmailAddress](verifiedEmailUrl(eori)).map(Option(_)).recover {
```
glidester, line 33: Out of curiosity, what is the significance of this change?

glidester (approved):
LGTM

app/uk/gov/hmrc/exports/connectors/CustomsDataStoreConnector.scala
MaciejRewera: Actually, I realized it can be simplified with `http.GET[Option[VerifiedEmailAddress]]`. I just wanted to avoid having empty `Some`. But I am not sure if it might happen.

udinobi: It cannot happen. The HttpReads raises an exception if it cannot instantiate the class. IMO the `Option` here is not good practice... practically you are saying that the library could return `null`.

udinobi: If you replace the placeholder `_` with a proper identifier you will notice that its type is `VerifiedEmailAddress`, which of course cannot be `null`.

test/it/uk/gov/hmrc/exports/connector/CustomsDataStoreConnectorSpec.scala
```diff
@@ -18,37 +18,72 @@ package uk.gov.hmrc.exports.connector
...
+import uk.gov.hmrc.exports.config.AppConfig
 import uk.gov.hmrc.exports.connectors.CustomsDataStoreConnector
 import uk.gov.hmrc.exports.models.VerifiedEmailAddress
-import uk.gov.hmrc.http.HttpClient
 
 class CustomsDataStoreConnectorSpec extends IntegrationTestSpec {
 
-  implicit val appConfig: AppConfig = inject[AppConfig]
+  implicit private val appConfig: AppConfig = inject[AppConfig]
+  private val connector = inject[CustomsDataStoreConnector]
```
udinobi, line 32: Not sure, but I think you will have two different `AppConfig` instances here.

udinobi (approved):

test/it/uk/gov/hmrc/exports/connector/CustomsDataStoreConnectorSpec.scala
MaciejRewera: Since it is injected, it depends on how `AppConfig` is defined. But I had a look and it has `@Singleton` annotation, so there should be just a sinigle instance.

PR #254, opened 2021-02-19
CEDS-2975 Trigger email sending for notifications with DMSDOC responses

udinobi:


app/uk/gov/hmrc/exports/connectors/CustomsDataStoreConnector.scala
```diff
@@ -16,23 +16,27 @@
...
+@Singleton
+class CustomsDataStoreConnector @Inject()(http: HttpClient)(implicit appConfig: AppConfig, ec: ExecutionContext) extends Logging {
 
   import CustomsDataStoreConnector._
 
   def getEmailAddress(eori: String)(implicit hc: HeaderCarrier): Future[Option[VerifiedEmailAddress]] =
-    http.GET[Option[VerifiedEmailAddress]](verifiedEmailUrl(eori)).recover {
-      case Upstream4xxResponse(_, NOT_FOUND, _, _) => None
+    http.GET[Option[VerifiedEmailAddress]](verifiedEmailUrl(eori)).recoverWith {
+      case Upstream4xxResponse(_, NOT_FOUND, _, _) => Future.successful(None)
```
glidester, line 36: I don't think this exception will ever be thrown for a 404 (just a None returned from the GET call) as per: [Tech Space Blog](https://confluence.tools.tax.service.gov.uk/pages/viewpage.action?spaceKey=TEC&postingDay=2020%2F6%2F30&title=Http-verbs+update)

app/uk/gov/hmrc/exports/services/notifications/NotificationHelper.scala
```diff
@@ -0,0 +1,120 @@
...
+    val emailParameters = EmailParameters(Map(EmailParameter.MRN -> mrn))
+    val sendEmailRequest = SendEmailRequest(List(emailAddress.address), TemplateId.DMSDOC_NOTIFICATION, emailParameters)
+    emailConnector.sendEmail(sendEmailRequest).map {
+      case EmailAccepted                      =>
+      case BadEmailRequest(message)           => logSendEmailError("Bad request", message, eori, mrn)
+      case InternalEmailServiceError(message) => logSendEmailError("Email service error", message, eori, mrn)
+    }
+  }
+
+  private def sendEmailIfDmsdocNotification(details: NotificationDetails, submission: Submission)(implicit hc: HeaderCarrier): Boolean = {
```
glidester, line 65: Does this need to return a Boolean if it will always be true? Probably better to return Unit

app/uk/gov/hmrc/exports/services/notifications/NotificationHelper.scala
```diff
@@ -0,0 +1,120 @@
...
+  private def saveParsedNotifications(notification: Notification, parsedDetails: Seq[NotificationDetails])(implicit hc: HeaderCarrier): Unit =
+    Future.sequence {
+      parsedDetails.map { details =>
+        saveParsedNotification(Notification(notification.actionId, notification.payload, Some(details)))
+      }
+    } map { outcomes =>
+      // IS THIS OK?
+      // Removing from the repo the unparsed notification only if all notifications
+      // derived from the former after the parsing were successfully saved.
+      if (outcomes.forall(identity)) notificationRepository.removeUnparsedNotificationsForActionId(notification.actionId)
```
glidester, line 88: :thinking: Given the scenario where a Notification contained multiple details, but one of the details fails to be parsed, would this not result in the Notification being re-parsed on each app startup, and so the successful details being inserted into Mongo again and again (duplicated)?

udinobi: Indeed. On the other side, removing no matter what the "un-parsed" notification will eliminate the chance to re-parse it if one of the detail response is not successfully saved.
It's not that I'm in favor of one or the other solution. The current implementation is just to bring up the issue and discuss it.

MaciejRewera: Current parser works in a way that if there is parsing error, no `NotificationDetails` are being returned. Instead, it throws exception so it's all-or-nothing approach. 
Therefore, there is no risk of having duplicated Notifications because of that.

udinobi: :thinking: The problem of the removal of the un-parsed notification is not related to the parsing outcome, isn't it? But to the outcome of storing the single/multiple parsed notifications as well as if the related Submission can be updated or not.

app/uk/gov/hmrc/exports/models/declaration/notifications/Notification.scala
```diff
@@ -42,6 +44,9 @@ case class Notification(actionId: String, payload: String, details: Option[Notif
 
 object Notification {
 
+  def unparsed(actionId: String, notificationXml: NodeSeq): Notification =
+    Notification(actionId = actionId, payload = notificationXml.toString, details = None)
```
MaciejRewera, line 48: 👍

PR #255, opened 2021-02-25
CEDS-2975 Trigger email sending

MaciejRewera:
This PR implements logic necessary to obtain required data
and sending request to `email` service.
Additionally, it also integrates this process into Notification
processing "pipeline" as an action executed after successful
parsing, saving and MRN update to the related Submission.

This operation is implemented in a way that it is completely
independent from any other process.

Majority of the email sending logic was created by @udinobi .

### DIFF
--- a/app/uk/gov/hmrc/exports/services/notifications/NotificationFactory.scala
+++ b/app/uk/gov/hmrc/exports/services/notifications/NotificationFactory.scala
@@ -17,38 +17,28 @@
 package uk.gov.hmrc.exports.services.notifications
 
 import javax.inject.Inject
-import play.api.Logger
+import play.api.Logging
 import uk.gov.hmrc.exports.models.declaration.notifications.Notification
 
 import scala.util.{Failure, Success, Try}
-import scala.xml.{NodeSeq, XML}
+import scala.xml.XML
 
-class NotificationFactory @Inject()(notificationParser: NotificationParser) {
-
-  private val logger = Logger(this.getClass)
+class NotificationFactory @Inject()(notificationParser: NotificationParser) extends Logging {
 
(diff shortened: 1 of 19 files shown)


app/uk/gov/hmrc/exports/services/notifications/NotificationFactory.scala
```diff
@@ -17,38 +17,28 @@
...
-import scala.xml.{NodeSeq, XML}
+import scala.xml.XML
 
-class NotificationFactory @Inject()(notificationParser: NotificationParser) {
-
-  private val logger = Logger(this.getClass)
+class NotificationFactory @Inject()(notificationParser: NotificationParser) extends Logging {
 
   def buildNotifications(actionId: String, notificationXml: String): Seq[Notification] =
     Try(XML.loadString(notificationXml)).map(notificationParser.parse) match {
```
glidester, line 29: `notificationParser.parse` can throw an exception if the xml is valid but does not contain the required elements. In this scenario `logParseExceptionAtPagerDutyLevel` will not be called (and I think it should).

glidester (changes requested):

app/uk/gov/hmrc/exports/services/notifications/NotificationFactory.scala
MaciejRewera: There is a case for `Failure(exc)` which is executed when either `XML.loadString` or `notificationParser.parse` throws any exception.
Also, there are tests in `NotificationFactory` (near the end of the file) that cover the scenario when NotificationParser throws an exception. Although it is not tested if logger is called, empty sequence is returned in case of a `Failure`. And there is a test that asserts against return value.

glidester (approved):
LGTM

PR #257, opened 2021-03-15
CEDS-3045 email resending

MaciejRewera:
This is a periodic job that sends email requests to email service.
It is based on SendEmailWorkItemRepository and SendEmailWorkItems.
Email sending process has been changed to unify treatment of email
triggers.

How email sending process looks like?

After parsing Notification, saving it into Mongo and updating related
Submission with MRN, SendEmailForDmsDocAction makes a check if this
Notification has DMSDOC status. If it has not, the process ends there.
However, if it is DMSDOC, then a new document is created in
sendEmailsWorkItems collection, containing Notification's document id
and MRN.
This process also works in a scenario where Notification arrived before
inserting Submission document into Mongo.

SendEmailsJob runs periodically with configured interval. It pulls
documents one by one from sendEmailsWorkItems that are not processed
yet, builds a request to email service, sends it and handles the result
of this call. There are 4 scenarios being handled:
1. Email Accepted - request has been accepted by email service. WorkItem
is marked as 'Succeeded'.
2. Missing Data - a piece of information required to build the request
cannot be obtained. WorkItem is marked as 'Failed'.
3. 4xx response from email service - potential problem on our side for
this particular request. WorkItem is marked as 'Failed'
4. 5xx response from email service - This usually means that there is no
reason to try sending subsequent requests. Because of that current
WorkItem is marked as 'Failed' but SendEmailsJob stops processing more
WorkItems. The next attempt will be taken with the next run.

SendEmailsJob is executed by ALL instances of this service, but using
WorkItemRepository ensures that given WorkItem is processed only
by a single instance during particular Job run.
An explanation for p. no. 4 above: during a single Job run there may be
as many requests sent as there are instances of customs-declare-exports
service before stopping the process for the rest of WorkItems.

Reasoning behind this design:

There was another proposition on the process. It assumed adding an
optional field to the existing Notification class which would contain
information necessary to handle email sending. 
However, after some trial and error it turned out that both solutions
are very similar in terms of complexity.
In the end, I decided to go this path because it allows to build
implementation with minimum interference into existing code and data
model. It also keeps email-related data separate from the actual
Notification data.
 I also wanted to avoid having all emails handled by a single instance.
In some cases this could lead to very unequal work distribution.


app/uk/gov/hmrc/exports/scheduler/jobs/SendEmailsJob.scala
```diff
@@ -0,0 +1,93 @@
...
+  override def interval: FiniteDuration = appConfig.sendEmailsJobInterval
+
+  override def execute(): Future[Unit] = {
+    logger.info("Starting SendEmailsJob execution...")
+
+    val now = DateTime.now()
+    val failedBefore = interval.minus(1.minute).toMinutes
+
+    def process(): Future[Unit] =
+      sendEmailWorkItemRepository.pullOutstanding(failedBefore = now.minusMinutes(failedBefore.toInt), availableBefore = now).flatMap {
```
glidester, line 64: According to the docs failedBefore means 'it will only consider WorkItems in FailedState if they were marked as Failed before the failedBefore dateTime'. Should this not be configurable (like the scheduled job run interval)? 

Just thinking about the scenario where if (for whatever reason) a particular email requests causes a 500 response from the email server (perhaps they have a bug and can not handle a particularly odd email address) but other emails could be sent. If on the first run attempt this email fails to be sent (and stops the scheduledJob processing other emails) it would be good to exclude this email from the next scheduledJob run to allow the other emails a chance to be sent. 

So setting this failedBefore to be greater than the scheduled job run interval would achieve this (by resulting in any failed email not being reattempted in the very next process run at least).

test/it/uk/gov/hmrc/exports/repositories/SubmissionRepositorySpec.scala
```diff
@@ -57,7 +59,7 @@ class SubmissionRepositorySpec extends IntegrationTestBaseSpec {
         val exc = repo.save(secondSubmission).failed.futureValue
 
         exc mustBe an[DatabaseException]
-        exc.getMessage must include("E11000 duplicate key error collection: customs-declare-exports.submissions index: actionIdIdx dup key")
+        exc.getMessage must include(s"E11000 duplicate key error collection: ${TestMongoDB.DatabaseName}.submissions index: actionIdIdx dup key")
```
glidester, line 62: :+1:

app/uk/gov/hmrc/exports/scheduler/jobs/SendEmailsJob.scala
MaciejRewera: For the scenario you described the Job would need to go through other emails as well in case of 500 response, actually. Otherwise you risk having 2 emails blocking the all others if they are picked up every second run.
Not sure if this may happen though. In the ticket it is specified that for 5xx response the Job should stop until next run time.

I will make it configurable, but I'll leave it's value and 5xx response handling as it is now. This requires wider discussion IMHO.

conf/application.conf
```diff
@@ -234,4 +234,7 @@ contexts {
     }
 }
 
-workItem.sendEmail.retryAfterMillis = 300000
+workItem.sendEmail {
+    retryAfterMillis = 300000
+    consideredFailedBefore = "4m"
+}
```
glidester, line 240: Missing new line, but apart from this look good.

glidester (changes requested):

glidester (approved):

PR #258, opened 2021-03-18
CEDS-3045 Pager Duty alert triggering when failing to send email

MaciejRewera:
Pager Duty alert triggering is split into 2 classes:
Manager and Validator.
Validator class is responsible only for determination
if PD alert should be triggered or not. 
Manager class collects all data required for Validator
and also it is responsible for triggereing the alert and
updating WorkItem document in Mongo with this 
information.

The split is also there to make it easier to test.

### DIFF
--- a/app/uk/gov/hmrc/exports/scheduler/jobs/emails/PagerDutyAlertManager.scala
+++ b/app/uk/gov/hmrc/exports/scheduler/jobs/emails/PagerDutyAlertManager.scala
@@ -17,19 +17,38 @@
 package uk.gov.hmrc.exports.scheduler.jobs.emails
 
 import javax.inject.Inject
+import play.api.Logging
+import reactivemongo.api.ReadPreference
+import reactivemongo.bson.BSONObjectID
 import uk.gov.hmrc.exports.config.AppConfig
 import uk.gov.hmrc.exports.models.emails.SendEmailDetails
-import uk.gov.hmrc.workitem.{Failed, WorkItem}
+import uk.gov.hmrc.exports.repositories.SendEmailWorkItemRepository
+import uk.gov.hmrc.workitem.WorkItem
 
-class PagerDutyAlertManager @Inject()(appConfig: AppConfig) {
+import scala.concurrent.{ExecutionContext, Future}
 
-  def isPagerDutyAlertRequiredFor(workItem: WorkItem[SendEmailDetails]): Boolean = {
-    val isStatusFailed = workItem.status == Failed
-    val isAlertNotTriggered = !workItem.item.alertTriggered
+class PagerDutyAlertManager @Inject()(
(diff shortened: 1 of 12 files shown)


app/uk/gov/hmrc/exports/scheduler/jobs/emails/PagerDutyAlertValidator.scala
```diff
@@ -0,0 +1,37 @@
...
+
+package uk.gov.hmrc.exports.scheduler.jobs.emails
+
+import javax.inject.Inject
+import play.api.Logging
+import uk.gov.hmrc.exports.config.AppConfig
+import uk.gov.hmrc.exports.models.emails.SendEmailDetails
+import uk.gov.hmrc.workitem.{Failed, WorkItem}
+
+class PagerDutyAlertValidator @Inject()(appConfig: AppConfig) extends Logging {
```
udinobi, line 25: Do you really need `Logging` ?

app/uk/gov/hmrc/exports/scheduler/jobs/emails/PagerDutyAlertManager.scala
```diff
@@ -0,0 +1,54 @@
...
+import reactivemongo.api.ReadPreference
+import reactivemongo.bson.BSONObjectID
+import uk.gov.hmrc.exports.config.AppConfig
+import uk.gov.hmrc.exports.models.emails.SendEmailDetails
+import uk.gov.hmrc.exports.repositories.SendEmailWorkItemRepository
+import uk.gov.hmrc.workitem.WorkItem
+
+import scala.concurrent.{ExecutionContext, Future}
+
+class PagerDutyAlertManager @Inject()(
```
udinobi, line 30: Cannot be a singleton?

app/uk/gov/hmrc/exports/scheduler/jobs/emails/PagerDutyAlertValidator.scala
MaciejRewera: Must have been left after some previous iteration.

app/uk/gov/hmrc/exports/scheduler/jobs/emails/PagerDutyAlertManager.scala
MaciejRewera: Could be. But I don't see much difference to be fair, but can make it Singleton if you think it is relevant.

udinobi: You're right here. I just think it's good practice. It does not prevent the approval anyway. Up to you at the end  :smile:

udinobi (approved):

PR #259, opened 2021-04-07
CEDS-3081 Stop email sending

MaciejRewera:
Email Notification for DMSDOC status is now
cancelled if there is a Notification received after
DMSDOC that contains one of the statuses:
- DMSCLE
- DMSINV
- DMSREJ

In case there is another DMSDOC received after one
of the 3 statuses above, email notification will be sent.

The 3 statuses above cancels any number of email 
notifications that were before.

### DIFF
--- a/app/uk/gov/hmrc/exports/models/declaration/notifications/ParsedNotification.scala
+++ b/app/uk/gov/hmrc/exports/models/declaration/notifications/ParsedNotification.scala
@@ -21,7 +21,13 @@ import play.api.libs.json.{Json, Reads, _}
 import reactivemongo.bson.BSONObjectID
 
 case class ParsedNotification(id: BSONObjectID = BSONObjectID.generate(), actionId: String, payload: String, details: NotificationDetails)
-    extends Notification
+    extends Notification with Ordered[ParsedNotification] {
+
+  override def compare(that: ParsedNotification): Int =
+    if (this.details.dateTimeIssued == that.details.dateTimeIssued) 0
+    else if (this.details.dateTimeIssued.isAfter(that.details.dateTimeIssued)) 1
+    else -1
+}
 
 object ParsedNotification {
 
--- a/app/uk/gov/hmrc/exports/models/declaration/notifications/ParsedNotification.scala
+++ b/app/uk/gov/hmrc/exports/models/declaration/notifications/ParsedNotification.scala
@@ -1,0 +1,63 @@
+/*
+ * Copyright 2021 HM Revenue & Customs
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package uk.gov.hmrc.exports.models.declaration.notifications
+
+import play.api.libs.functional.syntax._
+import play.api.libs.json.{Json, Reads, _}
+import reactivemongo.bson.BSONObjectID
+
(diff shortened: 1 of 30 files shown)


MaciejRewera:
Test this please

app/uk/gov/hmrc/exports/repositories/NotificationRepository.scala
```diff
@@ -43,18 +43,29 @@ class NotificationRepository @Inject()(mc: ReactiveMongoComponent)(implicit ec:
...
+    find("details" -> JsNull).map(toUnparsedNotifications)
 
   def removeUnparsedNotificationsForActionId(actionId: String): Future[WriteResult] =
     remove("actionId" -> JsString(actionId), "details" -> JsNull)
+
+  private def toParsedNotifications(notifications: Seq[Notification]): Seq[ParsedNotification] = notifications.collect {
+    case notification: ParsedNotification => notification
+  }
+
+  private def toUnparsedNotifications(notifications: Seq[Notification]): Seq[UnparsedNotification] = notifications.collect {
```
glidester, line 68: Nit: maybe change the names to begin with 'filter' or 'includeOnly' as the name 'toXXXXX' sounds like your are converting them into ParsedNotifications rather than just filtering out the unparsed ones.

app/uk/gov/hmrc/exports/scheduler/jobs/emails/EmailCancellationValidator.scala
```diff
@@ -0,0 +1,43 @@
...
+
+@Singleton
+private[emails] class EmailCancellationValidator @Inject()(notificationRepository: NotificationRepository) {
+
+  private val statusesCancellingEmailSending = Set(REJECTED, CLEARED, CANCELLED)
+
+  def isEmailSendingCancelled(sendEmailDetails: SendEmailDetails)(implicit ec: ExecutionContext): Future[Boolean] =
+    notificationRepository.findNotificationsByMrn(sendEmailDetails.mrn).map { notifications =>
+      val currentDmsDocNotification = notifications
+        .find(_.id == sendEmailDetails.notificationId)
```
glidester, line 34: This will find the first DMSDOC notification, but what about subsequent DMSDOC notifications (that might have been sent later)? Should you not get the last DMSDOC notification for this MRN?

MaciejRewera: The other DMSDOC would have it's own WorkItem created. This process relates to specific DMSDOC, not to all DMSDOCs for given MRN. So if one DMSDOC has been cancelled, there might still be another DMSDOC received for this MRN and an email should be sent. And the other DMSDOC will be checked separately.

app/uk/gov/hmrc/exports/models/declaration/notifications/ParsedNotification.scala
```diff
@@ -0,0 +1,69 @@
...
+ * limitations under the License.
+ */
+
+package uk.gov.hmrc.exports.models.declaration.notifications
+
+import play.api.libs.functional.syntax._
+import play.api.libs.json.{Json, Reads, _}
+import reactivemongo.bson.BSONObjectID
+
+case class ParsedNotification(id: BSONObjectID = BSONObjectID.generate(), actionId: String, payload: String, details: NotificationDetails)
```
glidester, line 23: Is there a good reason to keep NotificationDetails as a separate class instead of just merging the fields directly into the ParsedNotification class? Is 'payload' needed here (repetition of verbose textual data which at this stage is redundant)?

glidester (approved):
Two comments (neither show stoppers though)

app/uk/gov/hmrc/exports/models/declaration/notifications/ParsedNotification.scala
MaciejRewera: The payload field has to be there for now, because frontend service expects it (it is not optional). But I want to change the frontend model and remove this field here in the next PR.

As for the NotificationDetails, I consider all other fields as kind of metadata and `details` field is the relevant data. Besides, having flat structure is more susceptible to information noise. Right now the data is more structured and so extending the class with new fields does not mix purposes of these (whether it's data, or some helper piece of information field).

glidester (approved):
:+1:

PR #260, opened 2021-04-12
CEDS-3077 Hardcode, for ExportsDeclaration, "DispatchLocation" to "EX" or "CO"

udinobi:


app/uk/gov/hmrc/exports/models/declaration/ExportsDeclaration.scala
```diff
@@ -47,10 +50,37 @@ case class ExportsDeclaration(
...
+
+  private val countriesFor_CO_declType = List("Jersey", "Guernsey")
 
-    import play.api.libs.json._
+  def codeForDispatchLocation(declRequest: ExportsDeclarationRequest): String =
+    (for {
+      consignee <- declRequest.parties.consigneeDetails
+      address <- consignee.details.address
+      maybeCO <- if (countriesFor_CO_declType.contains(address.country)) Some(SpecialFiscalTerritory) else None
+    } yield maybeCO).getOrElse(OutsideEU)
```
MaciejRewera, line 81: I must say I was surprised by this solution and that this logic happens actually in controller.
I'd argue if this is a good place to put this logic, because if you imagine having more conditionality for declaration, this file would hold too much responsibilities.
For the time being, it seems also a bit inefficient, because you need to perform this logic on every update.

I'd suggest to postpone this logic to the model conversion process which happens only once in the Declaration's lifetime. You can find the responsible class [here](https://github.com/hmrc/customs-declare-exports/blob/d4227ed01c405c01d95614496b52299f46e6012d/app/uk/gov/hmrc/exports/services/mapping/declaration/TypeCodeBuilder.scala#L26).
The good thing is that you already have the whole `ExportsDeclaration` provided as parameter, so you'd just need to move this logic there.

app/uk/gov/hmrc/exports/models/declaration/ExportsDeclaration.scala
```diff
@@ -47,10 +50,37 @@ case class ExportsDeclaration(
 
 object ExportsDeclaration {
 
-  object REST {
+  def apply(id: String, eori: Eori, declarationRequest: ExportsDeclarationRequest): ExportsDeclaration =
```
MaciejRewera, line 53: I like the idea of reversing the responsibility of type conversion 👍 The code is more readable on client's side as well.

test/unit/uk/gov/hmrc/exports/services/email/EmailSenderSpec.scala
```diff
@@ -25,14 +28,9 @@ import uk.gov.hmrc.exports.connectors.{CustomsDataStoreConnector, EmailConnector
...
 import uk.gov.hmrc.exports.models.emails._
 import uk.gov.hmrc.exports.repositories.SubmissionRepository
-import uk.gov.hmrc.http.HeaderCarrier
-
-import scala.concurrent.ExecutionContext.Implicits.global
-import scala.concurrent.Future
 
 class EmailSenderSpec extends UnitSpec {
 
-  private implicit val hc: HeaderCarrier = mock[HeaderCarrier]
```
MaciejRewera, line 35: 👍

test/unit/uk/gov/hmrc/exports/services/notifications/NotificationServiceSpec.scala
```diff
@@ -36,14 +39,9 @@ import uk.gov.hmrc.exports.models.declaration.notifications.{NotificationDetails
...
 import uk.gov.hmrc.exports.repositories.{NotificationRepository, SubmissionRepository}
 import uk.gov.hmrc.exports.services.notifications.receiptactions.NotificationReceiptActionsExecutor
-import uk.gov.hmrc.http.HeaderCarrier
-
-import scala.concurrent.ExecutionContext.Implicits.global
-import scala.concurrent.Future
 
 class NotificationServiceSpec extends UnitSpec with IntegrationPatience {
 
-  private implicit val hc: HeaderCarrier = mock[HeaderCarrier]
```
MaciejRewera, line 46: 👍

MaciejRewera (changes requested):
The logic implementation looks good, but I disagree with the plecement of it.
I have mentioned my concerns in the comment above. Another thing to consider is that it would be safer to have this logic where I suggested, because right now you'd need to remember about it if you'd need to add another controller method.

Simply, I think this is executed too early and so it needs to be performed too often.

test/unit/uk/gov/hmrc/exports/controllers/request/ExportsDeclarationRequestSpec.scala
```diff
@@ -81,13 +81,13 @@ class ExportsDeclarationRequestSpec extends UnitSpec with ExportsDeclarationBuil
     natureOfTransaction = Some(natureOfTransaction)
   )
 
-  "Request" should {
+  "ExportsDeclarationRequest" should {
     "map to ExportsDeclaration" in {
-      request.toExportsDeclaration(id, Eori(eori)) mustBe declaration
+      ExportsDeclaration(id, Eori(eori), exportsDeclarationRequest) mustBe exportsDeclaration
```
MaciejRewera, line 86: This spec is for `ExportsDeclarationRequest`, but you are testing `ExportsDeclaration` here.

app/uk/gov/hmrc/exports/services/mapping/declaration/TypeCodeBuilder.scala
```diff
@@ -36,11 +38,18 @@ class TypeCodeBuilder @Inject()() extends ModifyingBuilder[ExportsDeclaration, D
...
+  private def createTypeCode(decType: AdditionalDeclarationType, dispatchLocation: String): DeclarationTypeCodeType = {
     val typeCodeType = new DeclarationTypeCodeType()
-    dispatchLocation.foreach { data =>
-      typeCodeType.setValue(data.dispatchLocation + decType.toString)
-    }
+    typeCodeType.setValue(dispatchLocation + decType.toString)
     typeCodeType
   }
+
+  private val countriesFor_CO_declType = List("GG", "JE", "Jersey", "Guernsey")
```
MaciejRewera, line 47: Why is there both country name and country code? We should be using either codes, or names.

test/unit/uk/gov/hmrc/exports/services/mapping/declaration/TypeCodeBuilderSpec.scala
```diff
@@ -16,29 +16,73 @@
...
+
+    s"set typeCode in Declaration to 'CO${declarationType}'" when {
+
+      "dispatchLocation in ExportsDeclaration is undefined and the destination country is 'GG'" in {
+        testTypeCode("CO", withDestinationCountry(Country(Some("GG"))) +: modifiers)
+      }
+
+      "dispatchLocation in ExportsDeclaration is 'EX' and the destination country is 'GG'" in {
+        testTypeCode("CO", withDispatchLocation("EX") +: withDestinationCountry(Country(Some("GG"))) +: modifiers)
+      }
```
MaciejRewera, line 56: Nit: To be entirely sure, you may have the same test case for other countries as well. I am not suggesting copy-pasting it over and over again. It would be more appropriate to for e.g. have a list of all countries and another list with `dispatchLocation`, and then use for-comprehension to run all combinations in a loop.

MaciejRewera (commented):
LGTM, the only reason why I have not approved yet is that I do not understand why there are both country names and codes in the `TypeCodeBuilder`

app/uk/gov/hmrc/exports/services/mapping/declaration/TypeCodeBuilder.scala
udinobi: Because I was not sure what values `locations.destinationCountry.code` can have, if codes only or countries only or both. If you know what are the possible values I have no problem of course to only include those.

test/unit/uk/gov/hmrc/exports/services/mapping/declaration/TypeCodeBuilderSpec.scala
udinobi: I was just following the pattern I have seen most used in the other tests here, where a unit test covers a specific case. No problem to group them. But I'll wait your confirmation about the possible values of `locations.destinationCountry.code`

app/uk/gov/hmrc/exports/services/mapping/declaration/TypeCodeBuilder.scala
MaciejRewera: We only store country codes. There was a confusion some time ago, because it turned out we store it inconsistently, but it is not resolved.

udinobi: :+1:

udinobi: I removed "Jersey" and "Guernsey" from the list and that's ok, but now I'm thinking that there was no harm in leaving in the list and we would have been future-proof. At the end what is relevant is if the destination country is "Jersey" or "Guernsey", by the code or by the name is not important.

MaciejRewera (approved):
LGTM
A note for some time later might be to take he country code values from the same file we use for FE service. But maybe it makes more sense to do once we start using the data store service.

PR #265, opened 2021-05-05
[CEDS-3168] Add new MUCR field to Data model

glidester:
Create new MUCR entity in the backend data model and
add it as an optional field to the ExportsDeclaration
entity.

If new field is present it is written as a previous
document element in the Dec XML model (similar to how
DUCRs are handled)


app/uk/gov/hmrc/exports/services/mapping/goodsshipment/PreviousDocumentsBuilder.scala
```diff
@@ -72,21 +72,30 @@ class PreviousDocumentsBuilder @Inject()() extends ModifyingBuilder[PreviousDocu
...
     val categoryCode = new PreviousDocumentCategoryCodeType()
     categoryCode.setValue("Z")
     previousDocument.setCategoryCode(categoryCode)
 
     val id = new PreviousDocumentIdentificationIDType()
-    id.setValue(ducr.ducr)
+    id.setValue(value)
     previousDocument.setID(id)
 
     previousDocument.setLineNumeric(new java.math.BigDecimal(1))
```
MaciejRewera, line 95: Should `LineNumeric` always be '1' ? If there are both DUCR and MUCR, then there are 2 elements with the same value. Should they not be different?

app/uk/gov/hmrc/exports/services/mapping/goodsshipment/PreviousDocumentsBuilder.scala
```diff
@@ -72,21 +72,30 @@ class PreviousDocumentsBuilder @Inject()() extends ModifyingBuilder[PreviousDocu
...
 
-  private def createDucrDocument(ducr: DUCR): GoodsShipment.PreviousDocument = {
+  private def createDucrDocument(ducr: DUCR): GoodsShipment.PreviousDocument = createPrevDoc(ducr.ducr, "DCR")
+
+  def buildThenAdd(mucr: MUCR, goodsShipment: GoodsShipment): Unit =
+    if (mucr.nonEmpty) {
+      goodsShipment.getPreviousDocument.add(createMucrDocument(mucr))
+    }
+
+  private def createMucrDocument(mucr: MUCR): GoodsShipment.PreviousDocument = createPrevDoc(mucr.mucr, "MCR")
```
MaciejRewera, line 82: I'd suggest extracting these 2 hard coded Strings into fields with descriptive names.

test/unit/uk/gov/hmrc/exports/services/mapping/goodsshipment/PreviousDocumentsBuilderSpec.scala
```diff
@@ -48,6 +48,69 @@ class PreviousDocumentsBuilderSpec extends UnitSpec with ExportsDeclarationBuild
...
+
+        val previousDocs = goodsShipment.getPreviousDocument
+        previousDocs.size must be(1)
+        previousDocs.get(0).getID.getValue must be(VALID_MUCR)
+        previousDocs.get(0).getCategoryCode.getValue must be("Z")
+        previousDocs.get(0).getTypeCode.getValue must be("MCR")
+        previousDocs.get(0).getLineNumeric must be(BigDecimal(1).bigDecimal)
+      }
+
+      "a MUCR has been specified along with other PreviousDocument elements" when {
```
MaciejRewera, line 65: Is it worth adding one more test case when there is MUCR and DUCR?

MaciejRewera (commented):
LGTM
Not approving yet due to question about `LineNumeric` field

app/uk/gov/hmrc/exports/services/mapping/goodsshipment/PreviousDocumentsBuilder.scala
glidester: I believe this line number relates to a line on the 'document' that you are referencing (so line 1 of this document has the relevant references for this declarations). Each previous document element is independent of each other.

test/unit/uk/gov/hmrc/exports/services/mapping/goodsshipment/PreviousDocumentsBuilderSpec.scala
glidester: Well my thinking is that a DUCR previous document element is no different to any other type of previous document element. There is no special status to a DUCR previous document element that if combined with a MUCR previous document element would cause the code any particular issue, so I'm not convinced of what value the test would add.

app/uk/gov/hmrc/exports/services/mapping/goodsshipment/PreviousDocumentsBuilder.scala
glidester: Just confirmed this with Liz too :+1:

app/uk/gov/hmrc/exports/services/mapping/goodsshipment/PreviousDocumentsBuilder.scala
glidester: :+1:

app/uk/gov/hmrc/exports/services/mapping/goodsshipment/PreviousDocumentsBuilder.scala
MaciejRewera: Thanks, I was missing the context.

test/unit/uk/gov/hmrc/exports/services/mapping/goodsshipment/PreviousDocumentsBuilderSpec.scala
MaciejRewera: It is a bit different, as the "entry point" is different - DUCR is taken from `ConsignmentReferences`.
The value of this test case would be to test one of the overloaded methods. 

I'll leave it to your opinion, but I'd add this test case to also cover all potential scenarios.

MaciejRewera (dismissed):

MaciejRewera (approved):

PR #266, opened 2021-05-10
CEDS-3167 metrics enhancement

MaciejRewera:


app/uk/gov/hmrc/exports/metrics/ExportsMetrics.scala
```diff
@@ -18,23 +18,133 @@ package uk.gov.hmrc.exports.metrics
...
+  }
+
+  object Timers {
+    val notificationTimer: Timer = Timer(notificationMetric)
+
+    val upstreamCustomsDeclarationsTimer: Timer = Timer(upstreamCustomsDeclarationsMetric)
+
+    val declarationFindAllTimer: Timer = Timer(declarationFindAllMetric)
+    val declarationFindSingleTimer: Timer = Timer(declarationFindSingleMetric)
+    val declarationUpdateTimer: Timer = Timer(declarationUpdateMetric)
```
glidester, line 132: To aid readability I'd suggest changing these from e.g. 'declarationFindAllMetric' (that sounds like 'find all metrics') to 'metricsFindAllDeclarations', 'metricsFindSingleDeclaration' & 'metricsUpdateDeclarations'

app/uk/gov/hmrc/exports/metrics/ExportsMetrics.scala
```diff
@@ -18,23 +18,133 @@ package uk.gov.hmrc.exports.metrics
...
+    val timer: Timer = Timer(path)
+    val callCounter: Counter = Counter(path)
+    val completionCounter: Counter = Counter(s"$path.success")
+    val failureCounter: Counter = Counter(s"$path.failed")
+  }
+
+  object Timers {
+    val notificationTimer: Timer = Timer(notificationMetric)
+
+    val upstreamCustomsDeclarationsTimer: Timer = Timer(upstreamCustomsDeclarationsMetric)
```
glidester, line 128: Maybe 'metricsUpstreamDeclarationSubmission' would be better to understand its purpose?

glidester (approved):
LGTM, just some Nits on naming as possible improvements on readability

app/uk/gov/hmrc/exports/metrics/ExportsMetrics.scala
MaciejRewera: I've spent quite a bit of time figuring out what naming convention to use. My thinking is that metrics relating to similar processes have their prefixes the same. Having 'metric' at the end sounds more natural to me, although I am aware English is not my first language ;) 

One solution I am considering is to create a structure of nested objects, similar to how files are organised in folders. I'll give it a try and see how it looks like.

glidester (approved):
Yes I really like the object style references, makes the intent of each metric much clearer. Good job :+1:

PR #268, opened 2021-05-19
[CEDS-3182] New transport payment option

glidester:
Adding in new payment option value that when
present should not be sent to DMS as part of
the XML payload.


MaciejRewera (approved):
LGTM

PR #277, opened 2021-07-01
CEDS-3254 Swap order of DocumentsProducedData fields for consistency

udinobi:


MaciejRewera (approved):

glidester (approved):
LGTM

test/unit/uk/gov/hmrc/exports/services/mapping/governmentagencygoodsitem/AdditionalDocumentsBuilderSpec.scala
```diff
@@ -50,14 +50,14 @@ class AdditionalDocumentsBuilderSpec extends UnitSpec with GovernmentAgencyGoods
       mappedDocuments.get(0).getSubmitter.getName.getValue mustBe issusingAuthorityName
 
       val writeoff = mappedDocuments.get(0).getWriteOff
-      writeoff.getAmountAmount mustBe null
+      Option(writeoff.getAmountAmount) mustBe None
```
MaciejRewera, line 53: Not sure if it makes a big difference and I agree that `null` should be avoided, but it has been used here due to the wco model being based on Java classes.
I am simplt not sure if this change may cause something to slip through?

app/uk/gov/hmrc/exports/migrations/changelogs/cache/RenameToAdditionalDocuments.scala
```diff
@@ -0,0 +1,86 @@
...
+
+  private val INDEX_ID = "id"
+  private val INDEX_EORI = "eori"
+
+  private val collectionName = "declarations"
+
+  override val migrationInformation: MigrationInformation =
+    MigrationInformation(
+      id = "CEDS-3254 Rename items.documentsProducedData to items.additionalDocuments",
+      order = 9,
```
MaciejRewera, line 43: `order` should actually be 3, because the last migration created was about splitting Notifications collection. The one with no. 8 can be removed in my opinion. It is a left-over from the previous implementation and was useful as a quick look-up for implementing new ones. But since we have the 3rd migration now, it serves no purpose anymore.

app/uk/gov/hmrc/exports/migrations/changelogs/cache/RenameToAdditionalDocuments.scala
```diff
@@ -0,0 +1,86 @@
...
+    }
+
+    logger.info(s"Applying '${migrationInformation.id}' db migration... Done.")
+  }
+
+  private def updateItem(item: Document): Document = {
+    val documentsProducedData = item.get("documentsProducedData", classOf[Document])
+    item.put("additionalDocuments", documentsProducedData)
+    item.remove("documentsProducedData")
+    new Document(item)
```
MaciejRewera, line 84: Small efficiency gain: you can simply return `item` here since it is mutable object.

app/uk/gov/hmrc/exports/migrations/changelogs/cache/RenameToAdditionalDocuments.scala
```diff
@@ -0,0 +1,86 @@
...
+      id = "CEDS-3254 Rename items.documentsProducedData to items.additionalDocuments",
+      order = 9,
+      author = "Lucio Biondi",
+      runAlways = true
+    )
+
+  override def migrationFunction(db: MongoDatabase): Unit = {
+    logger.info(s"Applying '${migrationInformation.id}' db migration...")
+
+    val batchSize = 20
```
MaciejRewera, line 51: Just double checking: have you tested if batches of this size will not cause memory issues?

MaciejRewera (changes requested):
Please chaneg the `order` value for migration. More explanation in the comment.

Also, it shows that we may need to think about a different way to order migrations.

app/uk/gov/hmrc/exports/migrations/changelogs/cache/RenameToAdditionalDocuments.scala
udinobi: :+1:

app/uk/gov/hmrc/exports/migrations/changelogs/cache/RenameToAdditionalDocuments.scala
udinobi: tbh I was wondering what was indeed the criteria used to select the right batch size. I tried 20, since I had in my local DB 25 declarations, and the migration completed successfully, at least locally.

test/unit/uk/gov/hmrc/exports/services/mapping/governmentagencygoodsitem/AdditionalDocumentsBuilderSpec.scala
udinobi: Not sure to understand the comment. If `writeoff.getAmountAmount` returns `null`, `Option` will take care of that and the value will be `None`. So no slip through. Isn't it?

MaciejRewera (approved):
LGTM

PR #283, opened 2021-07-21
CEDS0-3340 reverse mapping baseline

MaciejRewera:


app/uk/gov/hmrc/exports/services/reversemapping/declaration/ExportsDeclarationXmlParser.scala
```diff
@@ -0,0 +1,64 @@
...
+  def fromXml(xml: String): ExportsDeclaration = {
+    val declarationXml = scala.xml.XML.loadString(xml)
+
+    buildExportsDeclaration(declarationXml)
+  }
+
+  private def buildExportsDeclaration(declarationXml: NodeSeq): ExportsDeclaration =
+    ExportsDeclaration(
+      id = UUID.randomUUID().toString,
+      eori = "GB1234567890",
```
udinobi, line 44: Just to be sure... is this the assignment we really want here?

MaciejRewera: There needs to be a placeholder for now. Rules around this eori are not obvious. I am not even sure, if we will not need to take it from authentication instead.

udinobi (approved):

PR #284, opened 2021-07-21
CEDS-3340 Enhance unit tests for XML parsers

MaciejRewera:


app/uk/gov/hmrc/exports/models/declaration/AdditionalDeclarationType.scala
```diff
@@ -30,6 +32,12 @@ object AdditionalDeclarationType extends Enumeration {
       Writes(v => JsString(v.toString))
     )
 
+  def fromString(str: String): Try[AdditionalDeclarationType] = Try {
+    AdditionalDeclarationType.values
+      .find(_.toString == str)
+      .getOrElse(throw new IllegalArgumentException(s"$str is not a valid value for AdditionalDeclarationType"))
```
udinobi, line 38: Could not this be returning an `Either` instead of a `Try` in order to avoid to raise an exception?

app/uk/gov/hmrc/exports/services/reversemapping/declaration/items/ProcedureCodesParser.scala
```diff
@@ -0,0 +1,74 @@
...
+import uk.gov.hmrc.exports.services.reversemapping.declaration.{DeclarationXmlParser, XmlParsingException}
+
+import scala.xml.NodeSeq
+
+class ProcedureCodesParser extends DeclarationXmlParser[Option[ProcedureCodes]] {
+
+  override def parse(itemXml: NodeSeq): XmlParserResult[Option[ProcedureCodes]] = validateInput(itemXml).map { _ =>
+    val procedureCode: Option[String] = (itemXml \ GovernmentProcedure).find { governmentProcedureNode =>
+      (governmentProcedureNode \ PreviousCode).nonEmpty
+    }.flatMap { governmentProcedureNode =>
```
udinobi, line 32: Maybe moving the flatMap block to a proper `def`?

udinobi (approved):
A couple of up-to-view comments

app/uk/gov/hmrc/exports/models/declaration/AdditionalDeclarationType.scala
MaciejRewera: This is not part of the parsers, so it should not rely on `XmlParserResult`. And returning `Either` here seems to me a bit like over-engineering.

app/uk/gov/hmrc/exports/services/reversemapping/declaration/DeclarationXmlParser.scala
```diff
@@ -0,0 +1,41 @@
...
+trait DeclarationXmlParser[A] {
+  def parse(inputXml: NodeSeq): XmlParserResult[A]
+}
+
+object DeclarationXmlParser {
+  type XmlParserResult[A] = Either[XmlParsingException, A]
+
+  implicit class EitherList[T](list: Seq[XmlParserResult[T]]) {
+    def toEitherOfList: XmlParserResult[Seq[T]] = {
+      def insideOut(rest: Seq[XmlParserResult[T]], acc: Seq[T]): XmlParserResult[Seq[T]] = rest match {
```
glidester, line 32: Nit: could make this a tail-recursive method

app/uk/gov/hmrc/exports/services/reversemapping/declaration/XmlParsingException.scala
```diff
@@ -16,8 +16,4 @@
 
 package uk.gov.hmrc.exports.services.reversemapping.declaration
 
-import scala.xml.NodeSeq
-
-trait XmlParser[A] {
-  def parse(inputXml: NodeSeq): A
-}
+case class XmlParsingException(message: String) extends RuntimeException(message)
```
glidester, line 19: Might be nice to add the xml that the exception was thrown when parsing

app/uk/gov/hmrc/exports/services/reversemapping/declaration/items/ProcedureCodesParser.scala
```diff
@@ -0,0 +1,74 @@
...
+    val errorMessages = Map(
+      isThereGovernmentProcedureWithPreviousCodeOnly -> "GovernmentProcedure element does not contain CurrentCode element",
+      areThereMultipleGovernmentProceduresWithPreviousCodes -> "There are multiple GovernmentProcedure elements with PreviousCode elements"
+    )
+
+    val errors = errorMessages.filter { case (validationFunction, _) => validationFunction(itemXml) }.values
+
+    if (errors.nonEmpty) Left(XmlParsingException(errors.mkString("[", ", ", "}")))
+    else Right(itemXml)
+  }
```
glidester, line 73: Just wondering where this logic is defined (is it in the tickets Beth and Monica wrote or is this just something you knew from past experience)?

app/uk/gov/hmrc/exports/services/reversemapping/declaration/XmlParsingException.scala
MaciejRewera: But this will be visible in the logs and I am not sure if we can show the whole xml there.

app/uk/gov/hmrc/exports/services/reversemapping/declaration/items/ProcedureCodesParser.scala
MaciejRewera: This should be present in the tickets, but you can always have a look into wco builders. In this case it's [TypeCodeBuilder](https://github.com/hmrc/customs-declare-exports/blob/61e25a825e5fbf0c04a8805a7ecee0a0fa33d0f2/app/uk/gov/hmrc/exports/services/mapping/declaration/TypeCodeBuilder.scala). And personally, I prefer to look into the code, because this is how it really works, not some spreadsheet extract.

app/uk/gov/hmrc/exports/services/reversemapping/declaration/DeclarationXmlParser.scala
MaciejRewera: According to IJ it is. I know that the self-invocation is not in the last `case` statement, but perhaps the compiler does the work?

glidester: pretty sure you need the @tailrec annotation in scala to make compiler turn this into a tail-recursive function (and yeah, last call needs to be the insideOut call so maybe need to re-order case statement).

app/uk/gov/hmrc/exports/services/reversemapping/declaration/XmlParsingException.scala
glidester: Well maybe we can just add the section of the xml that failed? Would be useful IMO

MaciejRewera: I simply don't know, but we can pass the question to BAs.

app/uk/gov/hmrc/exports/services/reversemapping/declaration/DeclarationXmlParser.scala
MaciejRewera: @tailrec is just for compile-time checking if the method is indeed tail-recursive. And it seems it does not need to be the last case statement.

glidester (approved):

PR #289, opened 2021-08-24
CEDS-3348 DIS-XML mapping to Exports Declaration - group 1

udinobi:


app/uk/gov/hmrc/exports/services/reversemapping/DeclarationId.scala
```diff
@@ -14,6 +14,6 @@
  * limitations under the License.
  */
 
-package uk.gov.hmrc.exports.services.reversemapping.declaration
+package uk.gov.hmrc.exports.services.reversemapping
 
-case class XmlParsingException(message: String) extends RuntimeException(message)
+case class DeclarationId(eori: String, mrn: String)
```
glidester, line 19: Think this name is misleading. MRN is a declarationId, eori is a userId. Should be called 'context' or something like that imo.

app/uk/gov/hmrc/exports/services/reversemapping/declaration/ExportsDeclarationXmlParser.scala
```diff
@@ -16,49 +16,53 @@
...
         createdDateTime = Instant.now(),
         updatedDateTime = Instant.now(),
         sourceId = None,
-        `type` = DeclarationType.STANDARD,
+        `type` = declarationType,
         dispatchLocation = None,
         additionalDeclarationType = additionalDeclarationType,
-        consignmentReferences = consignmentReferences,
-        linkDucrToMucr = None,
+        consignmentReferences = consignmentReferences.map(_.copy(mrn = Some(declarationId.mrn))),
```
glidester, line 64: I don't think setting the MRN is actually part of this ticket (only `lrn` and `ducr` consignmentReferences fields are mentioned). Probably better to leave it as None.

udinobi: Problem is that the MRN is not part of any of the `XML mapping` tickets. Still I think it does not hurt to have it as reminder of what we are supposed to do if the data cannot be directly derived from parsing the xml.

udinobi: Actually the MRN we'll be sending to DIS is not the same one in ConsignmentReferences, so I'll leave it to None :+1:

glidester (approved):

PR #293, opened 2021-09-01
CEDS-3363 XML to Exports Declaration mapping - group 7

udinobi:


app/uk/gov/hmrc/exports/services/reversemapping/declaration/items/SingleItemParser.scala
```diff
@@ -17,19 +17,45 @@
...
+        fiscalInformation = parseFiscalInformation(domesticDutyTaxParties),
+        additionalFiscalReferencesData = parseAdditionalFiscalReferencesData(domesticDutyTaxParties),
+        statisticalValue = parseStatisticalValue(itemXml)
+      )
     }
+
+  private def parseAdditionalFiscalReferencesData(domesticDutyTaxParties: NodeSeq): Option[AdditionalFiscalReferences] = {
+    val references = (domesticDutyTaxParties \ ID).flatMap { reference =>
+      val text = reference.text
+      if (text.length > 1) Some(AdditionalFiscalReference(country = text.take(2), reference = text.drop(2))) else None
```
udinobi, line 50: Schema has `an..17` as format for `(domesticDutyTaxParties \ ID)`
I think no harm to have the condition `text.length > 1`, if not let me know.

MaciejRewera: Length check seems unnecessary in this place. Both `take` and `drop` methods return empty String should the index not exist.
And instead of calling simply `text` method on `reference`, I think it would be better to use `StringOption` and work with `Option[String]` type instead.

test/unit/uk/gov/hmrc/exports/services/reversemapping/declaration/items/SingleItemParserSpec.scala
```diff
@@ -33,7 +35,7 @@ class SingleItemParserSpec extends UnitSpec with EitherValues {
   override def beforeEach(): Unit = {
     super.beforeEach()
 
-    reset(procedureCodesParser)
+    when(procedureCodesParser.parse(any[NodeSeq])(any[MappingContext])).thenReturn(Right(None))
```
MaciejRewera, line 38: It is still worth having `reset` method here, to reset mock's counters for e.g.

test/unit/uk/gov/hmrc/exports/services/reversemapping/declaration/transport/TransportParserSpec.scala
```diff
@@ -86,23 +47,14 @@ class TransportParserSpec extends UnitSpec with EitherValues {
...
-      verify(meansOfTransportCrossingTheBorderNationalityParser).parse(any[NodeSeq])(any[MappingContext])
-      verify(meansOfTransportCrossingTheBorderTypeParser).parse(any[NodeSeq])(any[MappingContext])
-      verify(meansOfTransportOnDepartureIDNumberParser).parse(any[NodeSeq])(any[MappingContext])
-      verify(meansOfTransportOnDepartureTypeParser).parse(any[NodeSeq])(any[MappingContext])
-      verify(transportLeavingTheBorderParser).parse(any[NodeSeq])(any[MappingContext])
-      verify(transportPaymentParser).parse(any[NodeSeq])(any[MappingContext])
     }
 
     "return a Transport instance" when {
       "all sub-parsers return Right or a value" in {
```
MaciejRewera, line 53: Nit: There are no sub-parsers anymore

app/uk/gov/hmrc/exports/services/reversemapping/declaration/items/SingleItemParser.scala
```diff
@@ -17,19 +17,45 @@
...
+        id = UUID.randomUUID().toString,
+        sequenceId = (itemXml \ SequenceNumeric).text.toInt,
+        procedureCodes = procedureCodes,
+        fiscalInformation = parseFiscalInformation(domesticDutyTaxParties),
+        additionalFiscalReferencesData = parseAdditionalFiscalReferencesData(domesticDutyTaxParties),
+        statisticalValue = parseStatisticalValue(itemXml)
+      )
     }
+
+  private def parseAdditionalFiscalReferencesData(domesticDutyTaxParties: NodeSeq): Option[AdditionalFiscalReferences] = {
```
MaciejRewera, line 47: IMHO this part deserves its own parser. It makes sense not to create separate parsers for very simple cases, but here you have more logic inside and therefore the test suite is also bigger.

app/uk/gov/hmrc/exports/services/reversemapping/declaration/transport/TransportParser.scala
```diff
@@ -19,50 +19,55 @@ package uk.gov.hmrc.exports.services.reversemapping.declaration.transport
...
+    if (specificCircumstances == "A20") Some(YesNoAnswer(YesNoAnswers.yes)) else None
+  }
+
+  private def parseDepartureTransportMeans(inputXml: NodeSeq): NodeSeq =
+    inputXml \ Declaration \ GoodsShipment \ Consignment \ DepartureTransportMeans
+
+  private def parseTransportPayment(inputXml: NodeSeq): Option[TransportPayment] = {
+    val paymentMethod = (inputXml \ Declaration \ Consignment \ Freight \ PaymentMethodCode).text
+    if (paymentMethod.nonEmpty) Some(TransportPayment(paymentMethod)) else None
+  }
```
MaciejRewera, line 72: 👍

app/uk/gov/hmrc/exports/services/reversemapping/declaration/items/SingleItemParser.scala
udinobi: The test was to added to avoid to have a country with 1 character only, since the schema does not specify a minimal length.

MaciejRewera (commented):
LGTM apart from this one method in `SingleItemParser`. I think it should be placed in a separate class. I'd be glad to hear other's opinion on that.

It is a good idea to merge those simple/primitive parsers into their parent 👍

test/unit/uk/gov/hmrc/exports/services/reversemapping/declaration/items/SingleItemParserSpec.scala
udinobi: But `beforeEach` is executed **before** each test, and `when` replaces any stub for that callsite. I cannot see how the counters could be different when a test starts. Still it's not a big deal to add the reset.

test/unit/uk/gov/hmrc/exports/services/reversemapping/declaration/transport/TransportParserSpec.scala
udinobi: `TransportParser` still uses 1 sub-parser,  `containersParser`

app/uk/gov/hmrc/exports/services/reversemapping/declaration/items/SingleItemParser.scala
udinobi: My 2 cents... Personally I do not see the advantage to have a separate parser for this simple logic (I have seen other more complicated "internal methods" than this one :wink:). And more in general I'm against to tailor the implementation according to the tests. I would rather have an additional test class if required. However, always open to have a discussion of course.

app/uk/gov/hmrc/exports/services/reversemapping/declaration/DeclarationXmlParser.scala
```diff
@@ -44,4 +45,8 @@ object DeclarationXmlParser {
       insideOut(list, Nil)
     }
   }
+
+  implicit class OptionalString(nodeSeq: NodeSeq) {
+    def toStringOption: Option[String] = StringOption(nodeSeq.text)
+  }
```
glidester, line 51: Maybe I'm wrong, but won't this create a new OptionalString instance for every NodeSeq in scope (even for those NodeSeqs that do not represent strings)? If so seems a bit wasteful no?

app/uk/gov/hmrc/exports/services/reversemapping/declaration/items/SingleItemParser.scala
glidester: I'm happy for this to be a private method inside this parser. :+1:

app/uk/gov/hmrc/exports/services/reversemapping/declaration/DeclarationXmlParser.scala
udinobi: Do you mean at runtime? I don't think so. Only for the **types** in scope. Do not forget that the class is just a template at compile time.

test/unit/uk/gov/hmrc/exports/services/reversemapping/declaration/items/SingleItemParserSpec.scala
MaciejRewera: This would work as you describe if you'd create a new stub before each test case. However, the stub instance is shared between test cases and only its behaviour is changed/overriden with `when` method. And this method does not reset the state of the mock, which means counters are not set to zero before each test.

app/uk/gov/hmrc/exports/services/reversemapping/declaration/items/SingleItemParser.scala
```diff
@@ -46,8 +46,10 @@ class SingleItemParser @Inject()(procedureCodesParser: ProcedureCodesParser) ext
 
   private def parseAdditionalFiscalReferencesData(domesticDutyTaxParties: NodeSeq): Option[AdditionalFiscalReferences] = {
     val references = (domesticDutyTaxParties \ ID).flatMap { reference =>
-      val text = reference.text
-      if (text.length > 1) Some(AdditionalFiscalReference(country = text.take(2), reference = text.drop(2))) else None
+      reference.toStringOption.map { text =>
+        val country = if (text.length > 1) text.take(2) else s"$text?"
+        AdditionalFiscalReference(country = country, reference = text.drop(2))
```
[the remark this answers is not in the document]
udinobi: @bethany-kish What do you think of Maciej's proposal?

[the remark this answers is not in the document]
udinobi: Beth response:
>  just documented the potential issue in the "Xml to Exports Data Model mapping recognized issues" Google doc

glidester (approved):
LGTM

MaciejRewera (approved):

PR #297, opened 2021-09-08
[CEDS-3394] Authorizations include automatic EORI population

glidester:
Adding new EoriSource field to the DeclarationHolder entity.


app/uk/gov/hmrc/exports/models/declaration/EoriSource.scala
```diff
@@ -0,0 +1,34 @@
...
+
+sealed trait EoriSource
+
+object EoriSource {
+  case object UserEori extends EoriSource
+  case object OtherEori extends EoriSource
+
+  val values = Seq(UserEori, OtherEori)
+
+  lazy val lookupByValue: Map[String, EoriSource] = values.map(entry => entry.toString -> entry).toMap
```
udinobi, line 30: Nit: does it really need to be `lazy`? The implicit `reads` is not, so maybe also `lookupByValue` could not be `lazy`, isn't it?

udinobi (approved):

app/uk/gov/hmrc/exports/services/reversemapping/declaration/parties/DeclarationHolderParser.scala
```diff
@@ -37,6 +36,6 @@ class DeclarationHolderParser extends DeclarationXmlParser[Seq[DeclarationHolder
     val id = (inputXml \ ID).toStringOption
     val catCode = (inputXml \ CategoryCode).toStringOption
 
-    DeclarationHolder(catCode, id)
+    DeclarationHolder(catCode, id, id.map(_ => EoriSource.OtherEori))
```
MaciejRewera, line 39: I am wondering if this is enough to use single value for all parsed decs. I don't know enough details, if there would be a problem when EORI belonged to the other category?

glidester: No because there is nothing to stop a user select 'OtherEori' in the UI and then enter their own EORI number. The whole ticket's purpose is to save the user from having to type in their own EORI, but if they insist on typing it in we are not going to try and stop then :slightly_smiling_face:

glidester: ...also I did discuss this approach with Beth beforehand and she was happy with it :wink:

MaciejRewera (approved):
LGTM

PR #301, opened 2021-09-16
CEDS-3410 Remove redundant endpoints

MaciejRewera:
There are new endpoints with the exact same
functionality.

### DIFF
--- a/app/uk/gov/hmrc/exports/controllers/SubmissionController.scala
+++ b/app/uk/gov/hmrc/exports/controllers/SubmissionController.scala
@@ -34,7 +34,7 @@ class SubmissionController @Inject()(
 )(implicit executionContext: ExecutionContext)
     extends RESTController(cc) with JSONResponses {
 
-  def create(id: String): Action[Unit] = authenticator.authorisedAction(parse.empty) { implicit request =>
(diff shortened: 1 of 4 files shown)


app/uk/gov/hmrc/exports/controllers/SubmissionController.scala
```diff
@@ -34,7 +34,7 @@ class SubmissionController @Inject()(
 )(implicit executionContext: ExecutionContext)
     extends RESTController(cc) with JSONResponses {
 
-  def create(id: String): Action[Unit] = authenticator.authorisedAction(parse.empty) { implicit request =>
+  def create(id: String): Action[AnyContent] = authenticator.authorisedAction(parse.default) { implicit request =>
```
glidester, line 37: Out of curiosity, why you changed this to `AnyContent`?

glidester (approved):

app/uk/gov/hmrc/exports/controllers/SubmissionController.scala
MaciejRewera: Firstly - I noticed it makes difference for testing and you can't use the same patterns.
Secondly - I don't see any benefit from having `Action[Unit]` here. So it was easier for me to do it this way, than to figure out a new way for testing.

PR #303, opened 2021-09-17
[CEDS-3477] Bug when user tries to view dec without transport payment

glidester:
Fix bug where a declaration in the cache has a TransportPayment entity with
no 'paymentMethod' field value.

Created a DB migration job to remove any TransportPayment entity with
no 'paymentMethod' field value from declarations in cache.


MaciejRewera (approved):
LGTM

PR #305, opened 2021-09-23
[CEDS-3478] Normalise invoice amount and exchange rate values

glidester:
Strip commas from values if present and ensure values that start
or end with a period are properly handled (normalised).


MaciejRewera (approved):
LGTM, but it seems like you are repeating the same code in two places. Do you think it would be worth extracting this code and re-use it in these places?

PR #325, opened 2022-01-11
CEDS-3702 create new test only create & submit dec endpoint

glidester:


app/uk/gov/hmrc/exports/controllers/testonly/TestOnlyController.scala
```diff
@@ -0,0 +1,134 @@
...
+    implicit request =>
+      for {
+        declaration <- declarationRepository.create(createDeclaration())
+        notification <- saveNotification(createNotification(declaration))
+        maybeDmsDocNotification <- optionallySaveDmsDocNotification(declaration, notification)
+        _ <- submissionRepository.save(createSubmission(declaration, notification))
+      } yield {
+        val status = if (notification != maybeDmsDocNotification) "CREATED WITH DMSDOC" else "CREATED"
+        val conRef = declaration.consignmentReferences
+        Ok(s"EORI:${request.body.eori}, LRN:${conRef.map(_.lrn).getOrElse("")}, MRN:${conRef.flatMap(_.mrn).getOrElse("")}, $status")
```
udinobi, line 58: Not sure it is relevant or not as it's just a test endpoint, but could not this be a Json response? Something like...
```
Json.obj("eori" -> request.body.eori, ...)
```

app/uk/gov/hmrc/exports/controllers/testonly/TestOnlyController.scala
```diff
@@ -0,0 +1,134 @@
...
+import uk.gov.hmrc.exports.util.ExportsDeclarationBuilder
+
+import java.time.{LocalDateTime, ZoneId, ZonedDateTime}
+import java.util.UUID
+import javax.inject.{Inject, Singleton}
+import scala.concurrent.{ExecutionContext, Future}
+import scala.util.Random
+
+@Singleton
+class TestOnlyController @Inject()(
```
udinobi, line 37: I would give a specific name to this class, not `TestOnlyController`. This file is already quite big, and if in future we need to add an additional test endpoint, we do not want for sure to add its logic to this controller again.

app/uk/gov/hmrc/exports/controllers/testonly/TestOnlyController.scala
glidester: There is no requirement for this endpoint to return anything, I've only included this response for debug purposes really.

udinobi (approved):

PR #328, opened 2022-02-08
CEDS-3671 Use value of /departure-transport for /border-transport when the latter has no value

udinobi:


darrylmurphy (approved):

app/uk/gov/hmrc/exports/services/mapping/declaration/BorderTransportMeansBuilder.scala
```diff
@@ -25,37 +25,39 @@ import wco.datamodel.wco.declaration_ds.dms._2._
...
-      }
+    val hasDepartureTransport = transport.hasDepartureTransportDetails
+
+    if (hasTransportLeavingTheBorder || hasBorder || hasDepartureTransport) {
+      val transportMeans = new Declaration.BorderTransportMeans
+
+      if (hasTransportLeavingTheBorder) appendTransportLeavingTheBorder(transport, transportMeans)
+
+      if (hasBorder) appendBorderTransport(transport, transportMeans)
+      else if (hasDepartureTransport) appendDepartureTransport(transport, transportMeans)
```
glidester, line 40: This bit seems a little confusing as both `appendBorderTransport` and `appendDepartureTransport` actually sets both BorderTransport and DepartureTransport values when called (but this `if` makes it seems like they are mutually exclusive).

udinobi: They both set BorderTransport only, but `appendDepartureTransport` is called **only** if the user did not enter any value in BorderTransport (and in this case `appendBorderTransport` is not called). Note that is Builder is dedicated to BorderTransport only.

app/uk/gov/hmrc/exports/services/mapping/declaration/BorderTransportMeansBuilder.scala
```diff
@@ -25,37 +25,39 @@ import wco.datamodel.wco.declaration_ds.dms._2._
...
-    if (hasDeparture || hasBorder) {
-      if (hasDeparture) {
-        appendDepartureTransport(transport, transportMeans)
-      }
-      if (hasBorder) {
-        appendBorderTransport(transport, transportMeans)
-      }
+    val hasDepartureTransport = transport.hasDepartureTransportDetails
+
+    if (hasTransportLeavingTheBorder || hasBorder || hasDepartureTransport) {
```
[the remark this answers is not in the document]
udinobi: But you want exec line 42 only if one these values are defined.

app/uk/gov/hmrc/exports/services/mapping/declaration/BorderTransportMeansBuilder.scala
glidester: Ok, I see. Maybe `appendDepartureTransport` could be named `appendBorderTransportWithDepartureValue' to make this more obvious?

PR #332, opened 2022-02-18
[BAU] Fix payload building issues

glidester:
Fix two issues:
1) When empty string values sent builder was producing schema non-compliant XML (e.g. <ns3:ID/> & <ns3:IdentificationTypeCode/>)
2) When BorderTransport does not specify an ID & IdentificationTypeCode so we use the DepartureTransport values instead, we were not
populating the RegistrationNationalityCode element with the value specified in BorderTransport

### DIFF
--- a/app/uk/gov/hmrc/exports/models/declaration/Transport.scala
+++ b/app/uk/gov/hmrc/exports/models/declaration/Transport.scala
@@ -36,9 +36,8 @@ case class Transport(
   meansOfTransportCrossingTheBorderIDNumber: Option[String] = None
 ) {
   def hasBorderTransportDetails: Boolean =
-    meansOfTransportCrossingTheBorderIDNumber.isDefined ||
-      meansOfTransportCrossingTheBorderType.nonEmpty ||
-      meansOfTransportCrossingTheBorderNationality.nonEmpty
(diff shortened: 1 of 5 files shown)


app/uk/gov/hmrc/exports/models/declaration/Transport.scala
```diff
@@ -36,9 +36,8 @@ case class Transport(
   meansOfTransportCrossingTheBorderIDNumber: Option[String] = None
 ) {
   def hasBorderTransportDetails: Boolean =
-    meansOfTransportCrossingTheBorderIDNumber.isDefined ||
-      meansOfTransportCrossingTheBorderType.nonEmpty ||
-      meansOfTransportCrossingTheBorderNationality.nonEmpty
+    meansOfTransportCrossingTheBorderIDNumber.fold(false)(_.nonEmpty) &&
+      meansOfTransportCrossingTheBorderType.fold(false)(_.nonEmpty)
```
udinobi, line 40: Uhm... if I'm not wrong these can be reduced to
```
meansOfTransportCrossingTheBorderIDNumber.exists(_.nonEmpty) &&
  meansOfTransportCrossingTheBorderType.exists(_.nonEmpty)
```

udinobi (approved):

PR #334, opened 2022-03-15
Ceds 3784 optimise use of mongo store when submitting declarations

glidester:
Removed from sumbmissions collection: ‘eoriIdx’ & ‘updateTimeIdx’
Removed from notifications collection: ‘actionIdIdx’, ‘detailsDateTimeIssuedIdx’ &  'detailsMrnIdx' 

Added new 'actionId' field to SendEmailDetails entity (with DB migration of existing records).

Reworked the declaration submission process to pair down number of queries performed from 6 -> 2.
Reworked the parse notification process to only update the submission's 'mrn' field if it has not
already been set.

When parsing notifications ensure that if no associated submission can be updated
that the parsing of the notification is considered as a failure. This is to prevent
the unlikely scenario of a notification being received and parsed before the
associated submission record is inserted on declaration submission.

Schedule notification reattempt every 60 seconds (in case of failure scenario above).

### DIFF
--- a/app/uk/gov/hmrc/exports/routines/RoutineRunner.scala
+++ b/app/uk/gov/hmrc/exports/routines/RoutineRunner.scala
@@ -18,17 +18,20 @@ package uk.gov.hmrc.exports.routines
 
 import akka.actor.{ActorSystem, Cancellable}
 import play.api.inject.ApplicationLifecycle
+import uk.gov.hmrc.exports.config.AppConfig
 import uk.gov.hmrc.exports.migrations.MigrationRoutine
 
 import javax.inject.Inject
 import scala.concurrent.Future
 import scala.concurrent.duration._
+import scala.util.Random
 
 class RoutineRunner @Inject()(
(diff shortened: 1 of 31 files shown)


app/uk/gov/hmrc/exports/routines/RoutineRunner.scala
```diff
@@ -39,4 +42,11 @@ class RoutineRunner @Inject()(
   }
 
   applicationLifecycle.addStopHook(() => Future.successful(migrationTask.cancel()))
+
+  val randomInitalDelay: FiniteDuration = Random.nextInt(30).seconds
```
udinobi, line 46: In order to always have a non-zero delay, if worth, what about something...
```
  val start =  5   // or any other value > 0
  val end   = 30
  val randomInitialDelay: FiniteDuration = start + Random.nextInt((end - start) + 1).seconds

```

udinobi (approved):

PR #359, opened 2022-06-28
CEDS-3920 Include MRN as valid search parameter for submissions

magicaljellybeans:


app/uk/gov/hmrc/exports/models/declaration/submissions/SubmissionQueryParameters.scala
```diff
@@ -19,9 +19,14 @@ package uk.gov.hmrc.exports.models.declaration.submissions
...
 import play.api.mvc.QueryStringBindable
 
-case class SubmissionQueryParameters(uuid: Option[String] = None, ducr: Option[String] = None, lrn: Option[String] = None) {
-
-  def isEmpty: Boolean = uuid.isEmpty && ducr.isEmpty && lrn.isEmpty
+case class SubmissionQueryParameters(
+  uuid: Option[String] = None,
+  ducr: Option[String] = None,
+  lrn: Option[String] = None,
+  mrn: Option[String] = None
```
glidester, line 26: Where is this new parameter used Tom?

glidester (approved):
This is fine for now Tom but when making a new field query-able we need to consider performance implications (like is there an index on this field in the mongo collection to make lookups performant). In this case there isn't but we will be revisiting the cancellation pages in the upcoming CAC epic and we can optimise then.

glidester (changes requested):
I've just spotted something I missed before that needs changing. In the class `uk.gov.hmrc.exports.services.SubmissionService` the method `isSubmissionAlreadyCancelled` isn't smart enough to distinguish between a declaration that has previously requested to be cancelled and was rejected (so user should be able to re-request cancellation) and a declaration that requested to be cancelled and was successful (so user shouldn't be able to re-request cancellation). 

Would you mind adding this logic in for me now Tom. Thanks :+1:

glidester (approved):

PR #368, opened 2022-07-18
CEDS-3973 Send EORI and 'EXRR', if not selected by the user, for 'Arrived' declarations with 'GVMS' ports

udinobi:


test/unit/uk/gov/hmrc/exports/services/mapping/AuthorisationHoldersBuilderSpec.scala
```diff
@@ -48,41 +51,130 @@ class AuthorisationHoldersBuilderSpec extends UnitSpec with ExportsDeclarationBu
...
         declaration.getAuthorisationHolder mustBe empty
       }
     }
   }
 
+  "AuthorisationHolders" when {
+    "the ExportsDeclaration instance is of 'Arrived' type and" when {
+      val additionalType = withAdditionalDeclarationType(STANDARD_FRONTIER)
+
+      "the user didn't enter 'EXRR', 'MIB' or 'CSE' on /add-authorisation-required and" when {
```
magicaljellybeans, line 87: Is it worth adding a test where the user has added one of these codes and checking that EXRR does not get added to the payload twice?

magicaljellybeans (approved):
LGTM. Just one question.

test/unit/uk/gov/hmrc/exports/services/mapping/AuthorisationHoldersBuilderSpec.scala
udinobi: :+1: Good idea

magicaljellybeans (approved):

PR #379, opened 2022-08-12
Ceds 3906 purge ancient submissions

darrylmurphy:


glidester (commented):
Can you please add a test into the `PurgeAncientSubmissionsJobSpec`to test that declarations that are older than 180 but who's 'enhancedStatusLastUpdated' is not in one of the specified states which requires deletion are not in fact deleted. This is the one important scenario thats missing that we need to ensure :slightly_smiling_face:

glidester (changes requested):
In `PurgeAncientSubmissionsJobSpec` these values should be dynamically defined rather than hard coded:
`val enhancedStatusLastUpdatedOlderThan = "2021-08-02T13:20:06Z[UTC]"`(better to make it exactly -180 days from today so we are performing boundary testing rather than some value in the distant past)
`val enhancedStatusLastUpdatedRecent = "2022-08-02T13:20:06Z[UTC]"` (this value must always be less than 180 days from the present, but in 171 days from now it won't be and the test will start failing!)

app/uk/gov/hmrc/exports/scheduler/jobs/PurgeAncientSubmissionsJob.scala
```diff
@@ -47,12 +47,29 @@ class PurgeAncientSubmissionsJob @Inject() (
...
+    )
 
-  private val olderThanDate = lte(statusLastUpdated, expiryDate)
+  private def olderThanDate = lte(statusLastUpdated, Codecs.toBson(expiryDate))
+
+  println(">>>>>>>>>" + olderThanDate)
+
+  // 2022-02-13T14:45:38.259Z
+
+  // '2022-02-13T14:45:37.529Z
```
glidester, line 72: Left in some debug printlns

test/it/uk/gov/hmrc/exports/scheduler/jobs/PurgeAncientSubmissionsJobSpec.scala
```diff
@@ -1,22 +1,31 @@
...
-  val testJob: PurgeAncientSubmissionsJob = app.injector.instanceOf[PurgeAncientSubmissionsJob]
+  val testJob: PurgeAncientSubmissionsJob = instanceOf[PurgeAncientSubmissionsJob]
+
+  private val clock: Clock = instanceOf[AppConfig].clock
+
+  val enhancedStatusLastUpdatedOlderThan = testJob.expiryDate
+  val enhancedStatusLastUpdatedRecent = ZonedDateTime.now(clock).minusDays(1)
+
+  println(">>>>>>>>>" + enhancedStatusLastUpdatedOlderThan)
+  println(">>>>>>>>>" + enhancedStatusLastUpdatedRecent)
```
glidester, line 28: and here

glidester (approved):
Two small clean ups needed but sound apart from that

glidester (approved):

PR #417, opened 2023-02-01
Ceds 4446 refactor of actions

darrylmurphy:


app/uk/gov/hmrc/exports/models/declaration/submissions/Action.scala
```diff
@@ -18,17 +18,69 @@ package uk.gov.hmrc.exports.models.declaration.submissions
...
-case class Action(
-  id: String,
-  requestType: RequestType,
-  requestTimestamp: ZonedDateTime = ZonedDateTime.now(ZoneId.of("UTC")),
-  notifications: Option[Seq[NotificationSummary]] = None
-) {
+sealed trait Action {
+
+  val id: String
+  val notifications: Option[Seq[NotificationSummary]]
```
magicaljellybeans, line 29: Is there a reason that `decId` is not an optional string declared in the trait?

app/uk/gov/hmrc/exports/models/declaration/submissions/Action.scala
```diff
@@ -38,10 +90,50 @@ object Action {
   implicit val readLocalDateTimeFromString: Reads[ZonedDateTime] = implicitly[Reads[LocalDateTime]]
     .map(ZonedDateTime.of(_, ZoneId.of("UTC")))
 
-  implicit val writes = Json.writes[Action]
+  private val allWrites = (JsPath \ "id").write[String] and
```
magicaljellybeans, line 93: Curious as to why you need customs read/writes?

app/uk/gov/hmrc/exports/models/declaration/submissions/Action.scala
```diff
@@ -18,17 +18,69 @@ package uk.gov.hmrc.exports.models.declaration.submissions
...
+  decId: String,
+  versionNo: Int
+) extends Action
+
+object AmendmentAction {
+  def apply(id: String, declaration: ExportsDeclaration, submission: Submission) =
+    new AmendmentAction(id, decId = declaration.id, versionNo = submission.latestVersionNo + 1)
+}
+
+case class ExternalAmendmentAction(
```
magicaljellybeans, line 74: Wasn't there the idea that `decId` would get populated for this class after it's been received and parsed by some job? (can't quite remember now)

app/uk/gov/hmrc/exports/repositories/DeclarationRepository.scala
```diff
@@ -105,7 +105,7 @@ object DeclarationRepository {
     // ?? Used for dashboard ??
     IndexModel(ascending("eori", s"$meta.updatedDateTime", s"$meta.status"), IndexOptions().name("eoriAndUpdateTimeAndStatusIdx")),
     // Used for pulling draft declarations for user in date order
-    IndexModel(ascending("eori", s"$meta.createdDateTime", s"$meta.status"), IndexOptions().name("eoriAndCreateTimeAndStatusIdx")),
+    IndexModel(ascending("eori", s"$meta.createdDateTime", s"$meta.status"), IndexOptions().name("eoriAndCreateTimeAndStatusIdxx")),
```
magicaljellybeans, line 108: Is that a typo?

magicaljellybeans (commented):
Where does the `submission.latestVersionNo` get updated by the way, can't seem to find that or a test for it?

Looking pretty good though!

darrylmurphy:
There's no context for it at the moment. As we only iterate on the version for amendments, so far we're only dealing with submission and cancellation so there's no need for it

app/uk/gov/hmrc/exports/models/declaration/submissions/Action.scala
darrylmurphy: > Where does the `submission.latestVersionNo` get updated by the way, can't seem to find that or a test for it?
> 
> Looking pretty good though!

There's no context for it at the moment. As we only iterate on the version for amendments, so far we're only dealing with submission and cancellation so there's no need for it

darrylmurphy: > 

The `decId` isn't a property of all actions. 

The `decId` is a mandatory String in all actions that need it, and it does not exist in the actions that do not.

Rather than being an optional String because it either exists or it doesn't based on the request type

app/uk/gov/hmrc/exports/models/declaration/submissions/Action.scala
darrylmurphy: Good question, this needs clarity

app/uk/gov/hmrc/exports/models/declaration/submissions/Action.scala
darrylmurphy: To read/write from an record to any one of the actions based on requestType

app/uk/gov/hmrc/exports/repositories/DeclarationRepository.scala
darrylmurphy: Yep

app/uk/gov/hmrc/exports/models/declaration/submissions/Action.scala
darrylmurphy: The theory is that rather than having one action that has all different rules based on its requestType, we can read the requestType and better model an action accordingly

app/uk/gov/hmrc/exports/models/declaration/submissions/Action.scala
darrylmurphy: Indeed you're right, will make the change

app/uk/gov/hmrc/exports/models/declaration/submissions/Action.scala
```diff
@@ -18,17 +18,70 @@ package uk.gov.hmrc.exports.models.declaration.submissions
...
 
 import java.time.{LocalDateTime, ZoneId, ZonedDateTime}
 
-case class Action(
-  id: String,
-  requestType: RequestType,
-  requestTimestamp: ZonedDateTime = ZonedDateTime.now(ZoneId.of("UTC")),
-  notifications: Option[Seq[NotificationSummary]] = None
-) {
+sealed trait Action {
```
glidester, line 26: I do not understand the reason for turning the original Action case class into a trait and then created a lot of specialist implementation case classes for all the different RequestTypes. Apart from being able to hard code the `SubmissionAction` `val versionNo: Int = 1` there is no benefit, and consequently it leads to a lot of duplicate code (as all the other Action case class implementations are exactly the same as each other).

app/uk/gov/hmrc/exports/models/declaration/submissions/Action.scala
```diff
@@ -38,10 +91,50 @@ object Action {
   implicit val readLocalDateTimeFromString: Reads[ZonedDateTime] = implicitly[Reads[LocalDateTime]]
     .map(ZonedDateTime.of(_, ZoneId.of("UTC")))
 
-  implicit val writes = Json.writes[Action]
+  private val allWrites = (JsPath \ "id").write[String] and
```
glidester, line 94: All these different read/writes are another downside of the Action trait approach

app/uk/gov/hmrc/exports/models/declaration/submissions/Submission.scala
```diff
@@ -39,8 +39,21 @@ case class Submission(
 
 object Submission {
 
-  implicit val format = Json.format[Submission]
+  implicit def format = Json.format[Submission]
 
-  def apply(declaration: ExportsDeclaration, lrn: String, ducr: String, action: Action): Submission =
+  def apply(declaration: ExportsDeclaration, lrn: String, ducr: String, action: SubmissionAction): Submission =
     new Submission(declaration.id, declaration.eori, lrn, None, ducr, actions = List(action), latestDecId = declaration.id)
+  def apply(uuid: String, declaration: ExportsDeclaration, notificationSummary: NotificationSummary, action: SubmissionAction): Submission =
```
glidester, line 46: Where is this constructor used in the code?

app/uk/gov/hmrc/exports/repositories/UpdateSubmissionsTransactionalOps.scala
```diff
@@ -116,21 +125,17 @@ class UpdateSubmissionsTransactionalOps @Inject() (
...
     def prependNotificationSummary(accumulator: Seq[NotificationSummary], notification: ParsedNotification): Seq[NotificationSummary] =
       NotificationSummary(notification, existingActions, accumulator) +: accumulator
 
     // Parsed notifications need to be sorted (asc), by dateTimeIssued, due to the (ACCEPTED => GOODS_ARRIVED_MESSAGE) condition
     val notificationSummaries = notifications.sorted.foldLeft(seed)(prependNotificationSummary)
 
-    // The resulting notificationSummaries are sorted again, this time (dsc), since it's not sure
-    // if they are more recent or not of the (maybe) existing notificationSummaries of the action.
-    val actionWithAllNotificationSummaries = action.copy(notifications = Some(notificationSummaries.sorted.reverse))
-    (actionWithAllNotificationSummaries, notificationSummaries)
```
glidester, line 134: Why are we no longer sorting these notifications in descending order?

glidester (changes requested):

app/uk/gov/hmrc/exports/connectors/CustomsDeclarationsConnector.scala
```diff
@@ -51,7 +53,10 @@ class CustomsDeclarationsConnector @Inject() (appConfig: AppConfig, httpClient:
     }
 
   def submitCancellation(submission: Submission, xml: String)(implicit hc: HeaderCarrier): Future[String] = {
-    def actionId: String = submission.actions.find(_.requestType == SubmissionRequest).fold("Not a SubmissionRequest?")(_.id)
+    def actionId: String = submission.actions.find {
+      case _: SubmissionAction => true
+      case _                   => false
```
udinobi, line 58: What does this pattern matching provide more than the former version? :thinking:

app/uk/gov/hmrc/exports/controllers/SubmissionController.scala
```diff
@@ -79,8 +77,10 @@ class SubmissionController @Inject() (authenticator: Authenticator, submissionSe
       submission.latestEnhancedStatus match {
         case Some(EnhancedStatus.ERRORS) => false // If the submission failed then don't block reuse of LRN
         case _ =>
-          submission.actions
-            .filter(_.requestType == SubmissionRequest)
+          submission.actions.filter {
+            case _: SubmissionAction => true
+            case _                   => false
+          }
```
udinobi, line 83: Same here

app/uk/gov/hmrc/exports/migrations/changelogs/submission/AddActionFieldsForAmend.scala
```diff
@@ -0,0 +1,77 @@
...
+  private val latestVersionNo = "latestVersionNo"
+
+  override def migrationFunction(db: MongoDatabase): Unit = {
+
+    logger.info(s"Applying '${migrationInformation.id}' db migration...")
+
+    val submissionCollection = db.getCollection("submissions")
+
+    submissionCollection
+      .find(and(exists(latestDecId), exists(uuid), exists(latestVersionNo)))
```
udinobi, line 50: :thinking: Not sure this is correct. I think you should only work on the documents where the fields you have to add, that is `decId` and `versionNo`, are actually missing. The current filter instead does not filter out any document, because the fields you are using are all already existing.

app/uk/gov/hmrc/exports/migrations/changelogs/submission/AddActionFieldsForAmend.scala
```diff
@@ -0,0 +1,77 @@
...
+        val submissionId = document.getString(uuid)
+        val submissionLatestDecId = document.getString(latestDecId)
+        val submissionLatestVersionNo = document.getInteger(latestVersionNo)
+
+        submissionCollection.updateOne(
+          Filters.eq(uuid, submissionId),
+          combine(
+            set("actions.$[elem1].decId", submissionId),
+            set("actions.$[elem2].decId", submissionLatestDecId),
+            set("actions.$[elem2].versionNo", submissionLatestVersionNo)
```
udinobi, line 62: :thinking: Asking... Is correct to assume that a `CancellationRequest` is always, in the array of `Action`s, the 2nd element?
Maybe I'm saying something wrong because now I do not remember well, but isn't the 1st element of the `Action` array always the last-in-time action inserted? If thats the case, you could also have...
```
  set("actions.$[elem1].decId", submissionLatestDecId),
  set("actions.$[elem2].decId", submissionId),
```

app/uk/gov/hmrc/exports/models/declaration/submissions/Action.scala
udinobi: I agreed with Tim. It seems more an academic exercise.

app/uk/gov/hmrc/exports/migrations/changelogs/submission/AddActionFieldsForAmend.scala
darrylmurphy: The job needs these fields to populate the decId or versionNo, otherwise the record is useless

app/uk/gov/hmrc/exports/models/declaration/submissions/Action.scala
udinobi: May I suggest to have a look [at this PR](https://github.com/hmrc/customs-declare-exports/pull/421/files#diff-9906e4255892c4389b4066892891b3c1cf71f90ccb52c119ed0e6f7329e35ef5)?
 I'm also using for the migration an ad-hoc `reads` to add nested fields to the model, but the approach is different from what your are doing.

app/uk/gov/hmrc/exports/migrations/changelogs/submission/AddActionFieldsForAmend.scala
darrylmurphy: `elem2` is a given name, it could quite easily be `elemCancellation` or whatever

app/uk/gov/hmrc/exports/models/declaration/submissions/EnhancedStatus.scala
```diff
@@ -43,9 +43,10 @@ object EnhancedStatus extends Enumeration {
 
   private val mappingForCANCELLED = (actions: Seq[Action]) => {
     def isCancellationWithCustomsPositionGranted(action: Action): Boolean =
-      action.requestType == CancellationRequest &&
-        action.notifications.exists(_.exists(_.enhancedStatus == CUSTOMS_POSITION_GRANTED))
-
+      action match {
+        case CancellationAction(_, _, notifications, _, _) => notifications.exists(_.exists(_.enhancedStatus == CUSTOMS_POSITION_GRANTED))
+        case _                                             => false
+      }
```
udinobi, line 49: Is this really better? :raised_eyebrow:

app/uk/gov/hmrc/exports/models/declaration/submissions/Action.scala
darrylmurphy: It provides more than this.
Each action has different requirements according to the `requestType`, so on reading the `requestType` and assigning to an appropriate case class allows us to better define the Action. 
For example the `versionNo` of `SubmissionAction` will always be 1, furthermore an `ExternalAmendmentAction` will have an `decId: Option[String]` but all other Actions must have `decId: String`
It also helps when instantiating Actions, where a custom apply for CancellationAction can ensure that the `versionNo` will be equal to the `submission.latestVersionNo` whereas a custom apply for `AmendmentAction` will iterate on `submission.latestVersionNo`
By using traits, types, custom apply methods we can implement all those rules that appear complicated in the story
It can make things more readable too

udinobi: Yesterday I was tired and it was too late to continue with the review :woozy_face:
Sorry Darryl, but I don't see anything in the ticket requirements really convincing to justify this multiplication of case classes. According to your logic we should then have 10 different `ExportsDeclaration` case classes based on journey type and additional type.
I would agree with your approach if the structure of the classes would require different fields but that's not case here. The structure is unique for all different `requestType`.

app/uk/gov/hmrc/exports/connectors/CustomsDeclarationsConnector.scala
udinobi: This is over-engineered. Why replacing a simple boolean expression with pattern matching? I'm really against this.

darrylmurphy: If you read the comments for why we take advantage of a SubmissionAction rather than requestType == "SubmissionRequest" it might make more sense

This is just a consequence of that and the intention isn't about just replacing an if statement with pattern matching (although looking at the difference really isn't that big a deal)

app/uk/gov/hmrc/exports/migrations/changelogs/submission/AddActionFieldsForAmend.scala
udinobi: But those fields are mandatory. They are not optional, they are always there. The filter checking if they are existing is useless

udinobi: btw... nothing preventing the migration from working :hugs:

app/uk/gov/hmrc/exports/migrations/changelogs/submission/AddActionFieldsForAmend.scala
udinobi: I don't quite understand your answer here

app/uk/gov/hmrc/exports/migrations/changelogs/submission/AddActionFieldsForAmend.scala
darrylmurphy: Yes true when all fields are mandatory and guaranteed to be there it doesn't provide much of a filter, I'll make a change here

app/uk/gov/hmrc/exports/models/declaration/submissions/EnhancedStatus.scala
udinobi: I really hope you want to revert this, please.

app/uk/gov/hmrc/exports/migrations/changelogs/submission/AddActionFieldsForAmend.scala
darrylmurphy: I'll make a change to demonstrate

app/uk/gov/hmrc/exports/models/declaration/submissions/Submission.scala
```diff
@@ -39,8 +39,21 @@ case class Submission(
 
 object Submission {
 
-  implicit val format = Json.format[Submission]
+  implicit def format = Json.format[Submission]
```
udinobi, line 42: Why a `def`? Is it dynamic in some way?

app/uk/gov/hmrc/exports/services/SubmissionService.scala
```diff
@@ -155,9 +156,9 @@ class SubmissionService @Inject() (
     }
 
   private def isSubmissionAlreadyCancelled(submission: Submission): Boolean =
-    submission.actions.find(_.requestType == CancellationRequest) match {
-      case Some(action) => action.latestNotificationSummary.fold(false)(_.enhancedStatus == CUSTOMS_POSITION_GRANTED)
-      case _            => false
+    submission.actions.exists {
+      case c: CancellationAction => c.latestNotificationSummary.fold(false)(_.enhancedStatus == CUSTOMS_POSITION_GRANTED)
+      case _                     => false
```
udinobi, line 161: :+1:

app/uk/gov/hmrc/exports/models/declaration/submissions/Action.scala
darrylmurphy: The requirements were for:
An action with a requestType == SubmissionRequest to have decId == submission.uuid and a versionNo == 1
An action with a requestType == CancellationRequest to have decId == submission.latestDecId and a submission.latestVersionNo
An action with a requestType == AmendmentRequest to have decId == declaration.id and a submission.latestVersionNo + 1
An action with a requestType == ExternalAmendmentRequest to have an optional decId and a submission.latestVersionNo + 1

Each action hinges on its requestType and each has different requirements
This approach allows us to to enforce all these rules from the ticket... whats the problem?

app/uk/gov/hmrc/exports/models/declaration/submissions/Submission.scala
darrylmurphy: Maybe left over from development - will adjust now

app/uk/gov/hmrc/exports/migrations/changelogs/submission/AddActionFieldsForAmend.scala
```diff
@@ -0,0 +1,76 @@
...
+  private val latestDecId = "latestDecId"
+  private val latestVersionNo = "latestVersionNo"
+
+  override def migrationFunction(db: MongoDatabase): Unit = {
+
+    logger.info(s"Applying '${migrationInformation.id}' db migration...")
+
+    val submissionCollection = db.getCollection("submissions")
+
+    submissionCollection.find.asScala.map { document =>
```
glidester, line 48: Does this find not need to predicate that the fields do not already exist (to guard against the migration being re-run for already migrated records in the mongo)? 
e.g. `find(and(exists("actions.$[submissionItem].decId", false), exists("actions.$[submissionItem].versionNo", false)))`

app/uk/gov/hmrc/exports/migrations/changelogs/submission/AddActionFieldsForAmend.scala
```diff
@@ -0,0 +1,76 @@
...
+      val submissionId = document.getString(uuid)
+      val submissionLatestDecId = document.getString(latestDecId)
+      val submissionLatestVersionNo = document.getInteger(latestVersionNo)
+
+      submissionCollection.updateOne(
+        Filters.eq(uuid, submissionId),
+        combine(
+          set("actions.$[submissionItem].decId", submissionId),
```
glidester, line 56: We also now need `set("actions.$[submissionItem].versionNo", submissionLatestVersionNo)`

glidester (changes requested):
Just a few tweaks needed to the migration job after the changes made this morning :+1:

app/uk/gov/hmrc/exports/migrations/changelogs/submission/AddActionFieldsForAmend.scala
darrylmurphy: I'll add the filter with a test, but probably highlights the importance that it shouldn't be a job run more than once if we can rely on our data.

app/uk/gov/hmrc/exports/migrations/changelogs/submission/AddActionFieldsForAmend.scala
```diff
@@ -0,0 +1,78 @@
...
+  private val latestDecId = "latestDecId"
+  private val latestVersionNo = "latestVersionNo"
+
+  override def migrationFunction(db: MongoDatabase): Unit = {
+
+    logger.info(s"Applying '${migrationInformation.id}' db migration...")
+
+    val submissionCollection = db.getCollection("submissions")
+
+    submissionCollection.find(Filters.and(exists("versionNo", false), exists("decId", false))).asScala.map { document =>
```
glidester, line 49: This filter does not work as intended. I think it is looking for fields at the submission level (that are not there), where it should be looking for fields at the action level. 
I tried running this migration locally and it overwrote existing action records that already had these fields defined.

glidester (changes requested):

app/uk/gov/hmrc/exports/migrations/changelogs/submission/AddActionFieldsForAmend.scala
```diff
@@ -0,0 +1,91 @@
...
+  private val latestVersionNo = "latestVersionNo"
+
+  override def migrationFunction(db: MongoDatabase): Unit = {
+
+    logger.info(s"Applying '${migrationInformation.id}' db migration...")
+
+    val submissionCollection = db.getCollection("submissions")
+
+    val updated = submissionCollection
+      .find()
```
glidester, line 49: As you've removed the filter from this first lookup, each time this migration job is run it will retrieve every Submission record from mongo and run an individual update query on each (possibly very costly). 
Could you please use the 'elemMatch' filter instead to find any submission record that contains an action without a `versionNo` field, e.g.:
`.find(Filters.and(elemMatch("actions", exists("versionNo", false))))`
This will make the migration much faster and safer to run in scenarios where it has already been applied

glidester (changes requested):

glidester (dismissed):

glidester (approved):

PR #421, opened 2023-02-08
CEDS-4370 Add Sequence Numbers for RoutingCountries, Containers & Seals

udinobi:


app/uk/gov/hmrc/exports/migrations/changelogs/cache/AddDeclarationMetaEntity.scala
```diff
@@ -43,8 +43,7 @@ class AddDeclarationMetaEntity extends MigrationDefinition with Logging {
     MigrationInformation(
       id = s"CEDS-4371 Add DeclarationMeta entity. Move several declaration metadata fields inside it.",
       order = 16,
-      author = "Tom Robinson",
-      runAlways = false
```
magicaljellybeans, line 47: I think we want to leave these enabled to run until we deploy to the integration environment?

app/uk/gov/hmrc/exports/migrations/changelogs/submission/AddSubmissionFieldsForAmend.scala
```diff
@@ -34,7 +34,7 @@ class AddSubmissionFieldsForAmend extends MigrationDefinition with Logging {
       id = s"CEDS-4364 Add new fields to the Submission model for amended Declarations.",
       order = 17,
       author = "Lucio Biondi",
-      runAlways = true
+      runAlways = true // Can be removed (default is false) once this migration is deployed to production
```
magicaljellybeans, line 37: I think integration as well, because we haven't deployed there in a long time (but other migrations have also not been applied there I believe)

app/uk/gov/hmrc/exports/migrations/changelogs/cache/AddSequencingOfMultipleItems.scala
```diff
@@ -0,0 +1,199 @@
...
+        val declaration = Json.parse(document.toJson).as[ExportsDeclaration](AddSequencingOfMultipleItems.reads)
+        val maxSequenceIds = Map(
+          ContainerKey -> declaration.transport.containers.fold(0)(_.size),
+          RoutingCountryKey -> declaration.locations.routingCountries.size,
+          SealKey -> declaration.transport.containers.fold(0)(_.foldLeft(0)(_ + _.seals.size))
+        )
+        val meta = declaration.declarationMeta
+        val declarationWithSequenceIds = declaration.copy(declarationMeta = meta.copy(maxSequenceIds = maxSequenceIds))
+
+        val filter = and(equal("eori", declaration.eori), equal("id", declaration.id))
```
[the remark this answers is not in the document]
udinobi: Because the index is `eori+id`. The default field added by Mongo for indexing docs is `_id` which is a different field from our field `id`.

app/uk/gov/hmrc/exports/migrations/changelogs/submission/AddSubmissionFieldsForAmend.scala
udinobi: Yes, yes. I just added this comment to mark the fact that the running of the migration should be disabled only after being deployed to the all envs. I will change the comment to make this more evident :+1:

magicaljellybeans (dismissed):

magicaljellybeans (approved):

PR #425, opened 2023-02-24
CEDS-4445 Add sequence ids to 'PackageInformation' instances

udinobi:


app/uk/gov/hmrc/exports/migrations/changelogs/cache/AddSequenceIdsToPackageInformation.scala
```diff
@@ -0,0 +1,189 @@
...
+      .find(filter)
+      .batchSize(batchSize)
+      .asScala
+      .foldLeft(0) { (counter, document) =>
+        Try(Json.parse(document.toJson(jsonWriter)).as[ExportsDeclaration](AddSequenceIdsToPackageInformation.reads)) match {
+          case Failure(exc) =>
+            val id = document.get("_id")
+            val error = exc.getMessage
+            logger.error(s"Error parsing document with _id($id): $error")
+            counter + 1
```
glidester, line 67: Just occurred to me that these failures are actually about failure to read the json documents using these custom 'former' reader functions. If this migration is ever run in the future and records match the given filter, we might get read failures because the structure of the entities has since been altered :thinking:

glidester (approved):

app/uk/gov/hmrc/exports/migrations/changelogs/cache/AddSequenceIdsToPackageInformation.scala
udinobi: Let's postpone a discussion on this topic to the next dev catchup, as I have something to say on this, OK?

dhanist-gadhiraju (approved):

PR #430, opened 2023-03-03
CEDS-4379 Amend - XML - Implement amendment request XML payload builder

magicaljellybeans:


test/unit/uk/gov/hmrc/exports/controllers/AmendmentControllerSpec.scala
```diff
@@ -0,0 +1,87 @@
...
+      "return 409 CONFLICT" in {
+        val completeDec = aDeclaration(withStatus(COMPLETE))
+        when(mockSubmissionService.markCompleted(any(), any())).thenReturn(Future.successful(Some(completeDec)))
+
+        val result = controller.create(postRequest)
+
+        status(result) mustBe CONFLICT
+      }
+    }
+    "amendment is found in AMENDMENT_DRAFT state" should {
```
glidester, line 73: What happens if we get a declarationId with a status of DRAFT (rather than AMENDMENT_DRAFT)? Worth a test to check we refuse the request in that scenario :thinking:

glidester (approved):
Looks good, just one extra test that might be good to add

glidester (approved):

PR #433, opened 2023-03-07
Ceds 4381 fetch and save dec from dis

darrylmurphy:


glidester (changes requested):
Looks good. Just requesting addition of one extra test :+1:

glidester (approved):

PR #442, opened 2023-03-24
CEDS-4537 Ensure PENDING declarations are visible on the dashboard

udinobi:


app/uk/gov/hmrc/exports/migrations/changelogs/cache/AddDeclarationMetaEntity.scala
```diff
@@ -43,9 +43,7 @@ class AddDeclarationMetaEntity extends MigrationDefinition with Logging {
     MigrationInformation(
       id = s"CEDS-4371 Add DeclarationMeta entity. Move several declaration metadata fields inside it.",
       order = 16,
-      author = "Tom Robinson",
-      // Can be removed (default is false) once this migration is deployed to all envs (production & integration included)
```
magicaljellybeans, line 47: I'm a bit wary of removing this because I don't think we have deployed to integration yet. It means we'd have to set them all to run again if we ever did deploy to integration.

magicaljellybeans (approved):

PR #445, opened 2023-04-06
CEDS-4581 Add an endpoint for the retrieval of 'Action' instances

udinobi:


darrylmurphy (dismissed):

darrylmurphy (dismissed):

app/uk/gov/hmrc/exports/controllers/SubmissionController.scala
```diff
@@ -49,6 +49,13 @@ class SubmissionController @Inject() (authenticator: Authenticator, submissionSe
     }
   }
 
+  def findAction(actionId: String): Action[AnyContent] = authenticator.authorisedAction(parse.default) { _ =>
```
glidester, line 52: We are not linking the lookup of an action ID with the EORI passed in the authenticated header. This could lead to leaking other people's actions to a particular users!

Can we do something similar to looking up a submission by its ID please?

glidester (changes requested):

PR #446, opened 2023-04-12
Ceds 4390 process amend request notifications

darrylmurphy:


app/uk/gov/hmrc/exports/repositories/UpdateSubmissionsTransactionalOps.scala
```diff
@@ -76,87 +77,56 @@ class UpdateSubmissionsTransactionalOps @Inject() (
...
-          session,
-          actionId,
-          notifications.head.details.mrn,
-          notificationSummaries.head,
-          submission.actions.updated(index, actionWithAllNotificationSummaries)
-        )
+        updateSubmissionRequest(session, action, mrn, summary, updatedActions)
+
+      case AmendmentRequest =>
+        val submissionStatus = notifications.head.details.status
```
udinobi, line 99: This should be `enhancedStatus` to avoid confusion.

app/uk/gov/hmrc/exports/repositories/UpdateSubmissionsTransactionalOps.scala
```diff
@@ -76,87 +77,56 @@ class UpdateSubmissionsTransactionalOps @Inject() (
...
+
+        submission.actions
+          .find(_.requestType == SubmissionRequest)
+          .flatMap(_.latestNotificationSummary.map(_.enhancedStatus))
+          .map(updateAmendmentRequest(session, action, mrn, summary, updatedActions, submissionStatus, _))
+          .getOrElse(Future.successful(None))
```
udinobi, line 105: Uhm... I would log something here... before returning just `None`.

app/uk/gov/hmrc/exports/repositories/UpdateSubmissionsTransactionalOps.scala
```diff
@@ -167,24 +137,108 @@ class UpdateSubmissionsTransactionalOps @Inject() (
...
-    actionId: String,
+    action: Action,
+    mrn: String,
+    summary: NotificationSummary,
+    actions: Seq[Action],
+    submissionStatus: SubmissionStatus,
+    latestEnhancedStatus: EnhancedStatus
+  ): Future[Option[Submission]] = (for {
+    decId <- action.decId
+    update <- submissionStatus match {
```
udinobi, line 150: The two cases of this `match` do the same thing apart the different `latestEnhancedStatus`. Why not deriving it before calling `updateAmendmentRequest` in order to have a single `Json.obj(...` ?
In case `SubmissionStatus` is not `REJECTED`, `CUSTOMS_POSITION_DENIED` or `CUSTOMS_POSITION_GRANTED` you could even avoid to call the method. Just a suggestion btw.

app/uk/gov/hmrc/exports/repositories/UpdateSubmissionsTransactionalOps.scala
```diff
@@ -147,37 +150,23 @@ class UpdateSubmissionsTransactionalOps @Inject() (
...
-              "latestVersionNo" -> action.versionNo,
-              "mrn" -> mrn,
-              "latestEnhancedStatus" -> latestEnhancedStatus,
-              "enhancedStatusLastUpdated" -> summary.dateTimeIssued,
-              "actions" -> actions
-            )
-          )
-        )
-      case _ => None
+    latestEnhancedStatus <- submissionStatus match {
```
udinobi: Nit: this syntax is confusing me. Would you mind to move the `match` expression to a proper def/val? Something like...
```
val enhancedStatus = submissionStatus match {
      case REJECTED | CUSTOMS_POSITION_DENIED => Some(ON_HOLD)
      case CUSTOMS_POSITION_GRANTED => Some(latestEnhancedStatus)
      case _  => None
}
for {
  ...
```

udinobi (approved):

PR #452, opened 2023-04-27
CEDS-4633 Amend migration that is throwing exception so that all subm…

magicaljellybeans:
…issions are migrated

### DIFF
--- a/app/uk/gov/hmrc/exports/migrations/changelogs/submission/AddNotificationSummariesToSubmissions.scala
+++ b/app/uk/gov/hmrc/exports/migrations/changelogs/submission/AddNotificationSummariesToSubmissions.scala
@@ -19,14 +19,17 @@ package uk.gov.hmrc.exports.migrations.changelogs.submission
 import com.mongodb.client.result.UpdateResult
 import com.mongodb.client.{MongoCollection, MongoDatabase}
 import org.bson.Document
-import org.mongodb.scala.model.Filters.{and, equal, exists}
+import org.mongodb.scala.model.Filters
+import org.mongodb.scala.model.Filters.{and, elemMatch, equal, exists, or}
 import play.api.Logging
 import play.api.libs.json.Json
 import uk.gov.hmrc.exports.migrations.changelogs.{MigrationDefinition, MigrationInformation}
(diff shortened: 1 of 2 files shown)


udinobi (approved):

app/uk/gov/hmrc/exports/controllers/CustomHealthController.scala
```diff
@@ -0,0 +1,39 @@
...
+class CustomHealthController @Inject() (
+  migrationStatus: MigrationStatus,
+  configuration: Configuration,
+  environment: Environment,
+  controllerComponents: ControllerComponents
+) extends HealthController(configuration, environment, controllerComponents) {
+
+  override def ping: Action[AnyContent] =
+    if (migrationStatus.isSuccess) {
+      super.ping
```
glidester, line 35: Looking at the standard `HealthRoutes` class it has this comment "Audit/Auth/Logging filters rely on HandlerDef configuration (controller) in order to attach it to the request, we must extend GeneratedRouter and use the invoker tricks.."
Just worried that our directly calling `healthController.ping` won't be picked up by the filters :thinking:

conf/prod.routes
```diff
@@ -1,3 +1,6 @@
+# Declare our custom health check first to ensure the deployment fails if migrations fail
+GET     /ping/ping        uk.gov.hmrc.exports.controllers.CustomHealthController.ping()
```
glidester, line 2: Does defining this before the `health.Routes` override the standard one (I would have assumed you had to define it after to override, but I may be wrong) :thinking:

glidester (commented):
Just two concerns about the implementation. The approach looks good, just need to ensure we don't break any standard functionality along the way :grinning:

conf/prod.routes
magicaljellybeans: First route declared has priority https://www.playframework.com/documentation/2.8.x/ScalaRouting#Routing-priority

app/uk/gov/hmrc/exports/controllers/CustomHealthController.scala
magicaljellybeans: Hmm, perhaps I need a CustomHealthRoutes that invokes our version of ping too

udinobi (approved):

PR #458, opened 2023-05-11
CEDS-4722 Fix bug that was inserting the wrong country code into XML

magicaljellybeans:


glidester (approved):
Would be good to detail what the cause of the bug was and how this code fixes that. I assume it was use of the `contains` method rather than `==` (but would be good to specify this in commit message) :wink:

PR #460, opened 2023-05-16
CEDS-4730 Add fetching the latest Notification by using the Action identifier

udinobi:


app/uk/gov/hmrc/exports/services/notifications/NotificationService.scala
```diff
@@ -32,6 +32,9 @@ class NotificationService @Inject() (
   notificationReceiptActionsScheduler: NotificationReceiptActionsScheduler
 )(implicit executionContext: ExecutionContext) {
 
+  def find(actionId: String): Future[Option[ParsedNotification]] =
```
glidester, line 35: What is returned here if an Action has multiple notifications associated with it? :thinking:

udinobi: I see what you mean only now. Actually, for the failed/rejected amendment the endpoint should always work, but for other uses it could be confusing and leading to errors. I will review it :+1:

PR #466, opened 2023-05-26
Ceds 4785 correct and add missing pointers

magicaljellybeans:


app/uk/gov/hmrc/exports/services/mapping/ExportsPointerToWCOPointer.scala
```diff
@@ -38,9 +38,17 @@ class ExportsPointerToWCOPointer @Inject() (environment: Environment) {
     if (lines.size != allLines.size)
       throw new IOException(s"File $pointerFile is malformed. Expecting rows NOT starting with 'declaration.' and with 2 values separated by '|'")
 
+    // The following method is a temporary hack to handle pointers that come from the frontend as a sequence, but for reasons of expediency we are mapping to the WCO Pointer that is actually a parent of the sequenced elements.
+    // Example: There can be up to 99 taric or nact codes, but we will point to the Commodity element if any of these change.
+    // These are handled by removing the final sequence id so that this class will parse them.
+    def checkForManyToOnePointers(pointer: String): String =
```
glidester, line 44: Nit: rather than 'checkFor..' I think 'convert...' would be better as the method is not just checking for them, it is modifying them (splitting hairs but just makes it more readable imo) :slightly_smiling_face:

glidester (approved):

glidester (approved):

PR #474, opened 2023-06-07
Ceds 4790 wco pointer json

darrylmurphy:


darrylmurphy:
`exports-wco-mapping.csv` has been replaced with `exports-wco-mapping.json` where pointers are represented as JSON objects with properties frontend (`cds`) and `wco`.

A `Pointers` case class is used to deserialise the JSON with custom reads that maintain existing validation such as _nonEmpty_, not starting with _declaration._ and matching dollar signs.

There is also an option to read flags and comments attached to pointers, where each set of pointers can have a collection of (String, String) to represent flags and a collection of Strings representing comments.

All can be seen in `ExportsPointerToWCOPointerSpec`.

The usage of Pointers is contained purely to the method `ExportsPointerToWCOPointer.mapping`, which returns `Map[String, Seq[String]]` as before.

app/uk/gov/hmrc/exports/services/mapping/ExportsPointerToWCOPointer.scala
```diff
@@ -120,3 +105,61 @@ class ExportsPointerToWCOPointer @Inject() (environment: Environment) {
...
+
+    override def reads(json: JsValue): JsResult[Pointers] = {
+
+      val cds: String = (json \ "cds").as[String](cdsReads)
+      val wco: String = (json \ "wco").as[String](wcoReads)
+
+      val flags: Option[Seq[(String, String)]] = (json \ "flags").asOpt[Seq[(String, String)]](flagsReads)
+      val comments: Option[Seq[String]] = (json \ "comments").asOpt[Seq[String]]
+
+      if (convertManyToOnePointers(cds).count(_ == '$') == wco.count(_ == '$')) {
```
magicaljellybeans, line 158: The pointers referred to in the body of `convertManyToOnePointers` need a flag indicating they need to be processed in a certain way (ie having the sequenceId and last full stop removed). So we'll be removing the method but keeping the logic.

magicaljellybeans: Although, the main reason I had this logic, which was to make the  number of "$" chars match between CDS and WCO pointers, is not here anymore so there isn't actually a need to do this.

magicaljellybeans: So the pointers mentioned in `convertManyToOnePointers` will need comments indicating they are not one to one pointers in the json file, which I'll add above.

magicaljellybeans: But this can be removed 158-160.

magicaljellybeans: And 131- 139

conf/exports-wco-mapping.json
```diff
@@ -0,0 +1,402 @@
...
+  {
+    "cds" : "items.$1.commodityMeasure.netMass",
+    "wco" : "42A.67A.68A.$1.23A.65A.128"
+  },
+  {
+    "cds" : "items.$1.commodityMeasure.supplementaryUnits",
+    "wco" : "42A.67A.68A.$1.23A.65A.130"
+  },
+  {
+    "cds" : "items.$1.nactCode.$2",
```
magicaljellybeans, line 47: Can we add a comment for this item and the next one (taric) to the effect of "CDS pointer refers to one in a sequence of codes, but WCO pointer points to the parent level element"

conf/exports-wco-mapping.json
```diff
@@ -0,0 +1,402 @@
...
+  {
+    "cds" : "items.$1.packageInformation.$2.typesOfPackages",
+    "wco" : "42A.67A.68A.$1.93A.$2.141"
+  },
+  {
+    "cds" : "items.$1.procedureCodes.procedure.code",
+    "wco" : "42A.67A.68A.$1"
+  },
+  {
+    "cds" : "items.$1.procedureCodes.additionalProcedureCodes.$2",
```
magicaljellybeans, line 83: Same here "CDS pointer refers to one in a sequence of codes, but WCO pointer points to the parent level element"

PR #476, opened 2023-06-13
Ceds 4784 parsing work items retries

darrylmurphy:


app/uk/gov/hmrc/exports/scheduler/jobs/emails/SendEmailsJob.scala
```diff
@@ -97,20 +97,29 @@ class SendEmailsJob @Inject() (
...
+    val status =
+      if (workItem.failureCount >= appConfig.parsingWorkItemsRetryLimit) {
+        logger.error(s"[Email parsing error] parsing id:${workItem.id}' Cancelled! Max retries reached!")
+        Cancelled
+      } else {
+        Failed
+      }
+
+    sendEmailWorkItemRepository.markAs(workItem.id, status).flatMap { _ =>
       pagerDutyAlertManager.managePagerDutyAlert(workItem.id).map(_ => ())
```
glidester, line 122: This complicates matters, I didn't know/forgot that we are already sending PD alert log messages on failure to send an email. :thinking:  

Looks like it only sends the PD log message on the first failure.

The `PagerDutyAlertManager` seems to perform a db read to lookup on the same `workItem` instance that we already have and use to provide the `workItem.id`!. 

I know its not your code but can you refactor the `PagerDutyAlertManager.managePagerDutyAlert` method for us so that it is passed the `workItem` instance (so no additional db lookup is required). Thanks

app/uk/gov/hmrc/exports/scheduler/jobs/emails/SendEmailsJob.scala
```diff
@@ -97,20 +97,29 @@ class SendEmailsJob @Inject() (
...
+        failOrCancelSingle(workItem)
 
       case Success(BadEmailRequest(msg)) =>
         logger.warn(s"Email service returned Bad Request response: [$msg]")
-        failSingle(workItem)
+        failOrCancelSingle(workItem)
 
       case Success(InternalEmailServiceError(msg)) =>
         logger.warn(s"Email service returned Internal Service Error response: [$msg]")
         logger.warn(s"Will try again in ${interval.toMinutes} minutes")
```
glidester, line 108: This might no longer be true, if the failure count equals 9 then it won't be tried again!

app/uk/gov/hmrc/exports/scheduler/jobs/emails/SendEmailsJob.scala
```diff
@@ -97,20 +97,29 @@ class SendEmailsJob @Inject() (
...
-        failSingle(workItem)
+        failOrCancelSingle(workItem)
     }
 
-  private def failSingle(workItem: WorkItem[SendEmailDetails]): Future[Unit] =
-    sendEmailWorkItemRepository.markAs(workItem.id, Failed).flatMap { _ =>
+  private def failOrCancelSingle(workItem: WorkItem[SendEmailDetails]): Future[Unit] = {
```
glidester, line 112: Can we pass to this method the `log.warn` messages from the calling method above so that it can output all the log messages? 

Reason for this change is that atm the cause of the processing failure will appear on a different log line than the error msg (and at a different log level) so disconnected and possible hidden from the reader (by a filter on the log level in kibana) .

glidester (changes requested):
General approach is good, just some stuff to do with the email processing due to the existing `PagerDutyAlertManager`

app/uk/gov/hmrc/exports/scheduler/jobs/emails/SendEmailsJob.scala
```diff
@@ -97,20 +97,29 @@ class SendEmailsJob @Inject() (
...
+    val status =
+      if (workItem.failureCount >= appConfig.parsingWorkItemsRetryLimit) {
```
glidester, line 114: Won't this still reach 11 (if `parsingWorkItemsRetryLimit` is set to 10)?

darrylmurphy: If a workItem attempts are equal to 10 times, it will be set to Cancelled, which will work for those created from now and in the future
For those already existing where the attempts are well over the limit already, they will also be Cancelled here

Tested for in `NotificationReceiptActionsRunnerSpec.scala`

conf/application.conf
```diff
@@ -211,8 +210,9 @@ scheduler {
         interval = "1d"
     }
 
-    send-emails.interval = "5m"
+    send-emails.interval = "10h"
```
glidester, line 213: Should this not be 1h, then with the 10 retries it will hit 10h total

test/it/uk/gov/hmrc/exports/config/AppConfigISpec.scala
```diff
@@ -37,9 +37,8 @@ class AppConfigISpec extends IntegrationTestSpec {
   private val developerHubClientId = "customs-declare-exports"
   private val draftTimeToLive: FiniteDuration = 30.days
   private val purgeDraftDeclarations = JobConfig(LocalTime.of(23, 30), 1.day)
-  private val sendEmailsJobInterval: FiniteDuration = 5.minutes
+  private val sendEmailsJobInterval: FiniteDuration = 10.hours
```
glidester, line 40: 1.hours

glidester (changes requested):
Email resend interval should be 1h

glidester (approved):

PR #488, opened 2023-08-31
CEDS-5015 Cancellation of amendments

udinobi:


app/uk/gov/hmrc/exports/controllers/AmendmentController.scala
```diff
@@ -36,18 +36,29 @@ class AmendmentController @Inject() (
...
 
         case Some(declaration) if declaration.declarationMeta.status != AMENDMENT_DRAFT =>
-          Future.successful(Conflict(s"Attempted to submit declaration with status ${declaration.declarationMeta.status} as an amendment."))
+          val message = s"Attempted to submit declaration(${declarationId}) with status ${declaration.declarationMeta.status} as an amendment."
+          Future.successful(Conflict(message))
 
-        case Some(declaration) => submissionService.amend(request.eori, request.body, declaration).map(Ok(_))
+        case Some(declaration) =>
+          if (submissionAmendment.isCancellation) {
+            if (submissionAmendment.fieldPointers.size != 1 || !submissionAmendment.fieldPointers.head.isBlank) {
```
glidester, line 55: Why are we checking both `isCancellation` and fieldPointer is 1 or blank (should not isCacellation be enough)? :thinking:

udinobi: Yep, it should be enough. I was just too conservative probably. I can remove it btw. Not a big deal

glidester (approved):

PR #491, opened 2023-09-19
Ceds 5022 amendments routing countries

darrylmurphy:


app/uk/gov/hmrc/exports/services/SubmissionService.scala
```diff
@@ -343,4 +344,17 @@ class SubmissionService @Inject() (
...
+    val routingCountriesPredicate = "declaration.locations.routingCountries"
+    val consignmentPointer = "declaration.consignment"
+
+    val (matches, unmatches) = fieldPointers.partition(_.contains(routingCountriesPredicate))
+
+    if (matches.nonEmpty) {
+      logProgress(decId, s"Converted ${matches.length} Routing Country pointers to 1 Consignment pointer")
+      unmatches :+ consignmentPointer
+    } else fieldPointers
+  }
```
glidester, line 359: This is not needed. All we need to do is modify the `locations.routingCountries.$1"` entry in the mapping file to point to the WCO pointer for consignment

conf/exports-wco-mapping.json
```diff
@@ -181,6 +181,10 @@
     "cds" : "locations.routingCountries.$1",
     "wco" : "42A.28A.62A.$1"
   },
+  {
+    "cds" : "consignment",
+    "wco" : "42A.28A"
+  },
```
glidester, line 187: This is not needed. Just need to change the `wco` value above on line 182 to `"42A.28A"`

glidester (changes requested):

PR #502, opened 2023-12-13
CEDS-5093 Enable reverse sorting on 'enhancedStatusLastUpdated' while fetching Submissions

udinobi:


glidester (commented):
All looks good but I'm not seeing any additional tests to test the ordering param :thinking:

glidester (approved):

PR #508, opened 2024-02-01
CEDS-5446 Safer, hopefully, implementation of concurrent updates to a Submission document

udinobi:


app/uk/gov/hmrc/exports/repositories/SubmissionRepository.scala
```diff
@@ -46,22 +46,36 @@ class SubmissionRepository @Inject() (val mongoComponent: MongoComponent)(implic
...
 
   def addAction(uuid: String, action: Action): Future[Option[Submission]] = {
     val filter = Json.obj("uuid" -> uuid)
     val update = Json.obj("$addToSet" -> Json.obj("actions" -> Json.toJson(action)))
+
     findOneAndUpdate(filter, update)
   }
 
-  def updateAction(submissionId: String, actionId: String, decId: String): Future[Option[Submission]] =
+  def updateExternalAmendmentAction(uuid: String, actionId: String, declarationId: String): Future[Option[Submission]] = {
```
udinobi, line 60: This operation should possibly be reduced to a single `findOneAndUpdate`, in order to avoid conflicting concurrent updates to the document.

app/uk/gov/hmrc/exports/repositories/RepositoryOps.scala
```diff
@@ -38,6 +43,10 @@ trait RepositoryOps[T] {
 
   val collection: MongoCollection[T]
 
+  // Field, if any, of the collection used to annotate the time, a java.time.Instant,
+  // of the last update (only, not for document insertions and replacements).
+  val lastUpdate: Option[String] = None
```
glidester, line 48: Can I suggest you rename this val as 'lastUpdatedFieldName' to call out what it is holding (first glance I thought this field held a date/time value)

app/uk/gov/hmrc/exports/repositories/RepositoryOps.scala
```diff
@@ -232,6 +247,32 @@ trait RepositoryOps[T] {
...
+
+  private val readFromIsoString: Reads[Instant] =
+    implicitly[Reads[String]].map(s => Instant.from(ISO_INSTANT.parse(s)))
+
+  val instantReads: Reads[Instant] =
+    MongoJavatimeFormats.instantReads or
+      (__ \ "$date").read[Instant](readFromIsoString) or
+      __.read[Instant](readFromIsoString)
+
+  private def removeFinalZeroes(instant: Instant): String = {
```
glidester, line 260: Could use a comment to explain why the trailing zeros are being removed

app/uk/gov/hmrc/exports/repositories/SubmissionRepository.scala
glidester: Yes I agree should be one update. We shouldn't have a situation when the submission record is found but the action is missing so should be one atomic operation

app/uk/gov/hmrc/exports/repositories/UpdateSubmissionsTransactionalOps.scala
```diff
@@ -51,17 +53,33 @@ class UpdateSubmissionsTransactionalOps @Inject() (
 
   private lazy val nonTransactionalSession = mongoComponent.client.startSession().toFuture()
 
-  def updateSubmissionAndNotifications(actionId: String, notifications: Seq[ParsedNotification], submission: Submission): Future[Submission] =
+  // All update ops use 'enhancedStatusLastUpdated', loaded at transaction start, for "locking" the document under update.
```
glidester, line 56: Need to update this comment. It should  refer to new 'lastUpdated' field and not mention "locking" as that is not the new approach

glidester (approved):

PR #515, opened 2024-03-15
CEDS-5583 Add AssociatedSubmissionId to declaration meta data in Expo…

glidester:
…rts BE

### DIFF
--- a/test/it/uk/gov/hmrc/exports/migrations/changelogs/submission/AddAssociatedSubmissionToDeclarationsISpec.scala
+++ b/test/it/uk/gov/hmrc/exports/migrations/changelogs/submission/AddAssociatedSubmissionToDeclarationsISpec.scala
@@ -1,0 +1,280 @@
+/*
+ * Copyright 2023 HM Revenue & Customs
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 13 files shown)


test/it/uk/gov/hmrc/exports/migrations/changelogs/submission/AddAssociatedSubmissionToDeclarationsISpec.scala
```diff
@@ -0,0 +1,280 @@
...
+        setupSubmissions(submissionWithMultipleActions)
+        runTest(declarationBeforeMigration(DRAFT), declarationBeforeMigration(DRAFT))
+      }
+      "associatedSubmissionId already defined" in {
+        setupSubmissions(submissionWithSingleAction())
+        runTest(declarationAfterMigration(), declarationAfterMigration())
+      }
+    }
+
+    "Logs declaration not updated" when {
```
magicaljellybeans, line 64: Is some kind of log check occurring on these? Can't see anything?

magicaljellybeans (approved):

test/it/uk/gov/hmrc/exports/migrations/changelogs/submission/AddAssociatedSubmissionToDeclarationsISpec.scala
glidester: It was my initial intention to add some checks of the logging but after looking into how to do this with Play it was a non-trivial task to be able to verify log messages during test runs. So instead I manually checked the logs and left the tests in to check that the records were not updated if they ran into any issues

PR #522, opened 2024-04-11
CEDS-5697 Create resilient reads for TransportCountry class to enable…

magicaljellybeans:
… safe migration

### DIFF
--- a/app/uk/gov/hmrc/exports/models/declaration/TransportCountry.scala
+++ b/app/uk/gov/hmrc/exports/models/declaration/TransportCountry.scala
@@ -16,10 +16,17 @@
 
 package uk.gov.hmrc.exports.models.declaration
 
-import play.api.libs.json.{Json, OFormat}
+import play.api.libs.functional.syntax.toFunctionalBuilderOps
+import play.api.libs.json._
 
 case class TransportCountry(countryName: Option[String])
 
 object TransportCountry {
-  implicit val format: OFormat[TransportCountry] = Json.format[TransportCountry]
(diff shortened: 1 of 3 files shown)


app/uk/gov/hmrc/exports/models/declaration/TransportCountry.scala
```diff
@@ -16,10 +16,17 @@
...
-import play.api.libs.json.{Json, OFormat}
+import play.api.libs.functional.syntax.toFunctionalBuilderOps
+import play.api.libs.json._
 
 case class TransportCountry(countryName: Option[String])
 
 object TransportCountry {
-  implicit val format: OFormat[TransportCountry] = Json.format[TransportCountry]
+  // TODO remove resilient reads once 5606 migration has been deployed successfully
+  implicit val reads: Reads[TransportCountry] = (
```
udinobi, line 26: Do you actually need it as `implicit`? Wouldn't be better as `private`?

app/uk/gov/hmrc/exports/models/declaration/TransportCountry.scala
```diff
@@ -16,10 +16,17 @@
...
+    (__ \ "countryName").readNullable[String] and
+      (__ \ "countryCode").readNullable[String]
+  )((name, code) => TransportCountry(code.orElse(name)))
+  implicit val writes: Writes[TransportCountry] = Json.writes[TransportCountry]
```
udinobi, line 30: Do you actually need it as `implicit`? Wouldn't be better as `private`?

PR #531, opened 2024-04-19
CEDS-5606 Amend migration to include private country mappings with id…

magicaljellybeans:
…entified old values present in prod

### DIFF
--- a/app/uk/gov/hmrc/exports/migrations/changelogs/cache/RenameAndConvertBorderTransportNationalityField.scala
+++ b/app/uk/gov/hmrc/exports/migrations/changelogs/cache/RenameAndConvertBorderTransportNationalityField.scala
@@ -20,12 +20,13 @@ import com.mongodb.client.MongoDatabase
 import org.bson.Document
 import org.mongodb.scala.model.Filters.{and, equal, exists}
 import play.api.Logging
+import play.api.libs.json.{JsArray, JsString, Json}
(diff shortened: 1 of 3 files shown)


app/uk/gov/hmrc/exports/migrations/changelogs/cache/RenameAndConvertBorderTransportNationalityField.scala
```diff
@@ -88,3 +89,1062 @@ class RenameAndConvertBorderTransportNationalityField(countriesService: Countrie
     logger.info(s"Finished applying '${migrationInformation.id}' db migration.")
   }
 }
+
+object RenameAndConvertBorderTransportNationalityField {
+  private val countryMappings =
```
glidester, line 94: Ah, I was thinking something more along the lines of `val allCountriesPlus = countriesService.allCountries ++ ("Norway" -> "NO")` rather than copying the whole countries file into the migration job. However this will work just as well and we'll be removing the migration jobs soon anyhow so not going to sweat it :wink:

glidester (approved):

app/uk/gov/hmrc/exports/migrations/changelogs/cache/RenameAndConvertBorderTransportNationalityField.scala
magicaljellybeans: Oh yeah, that would have done it

PR #532, opened 2024-04-19
CEDS-5729: Filter out invalid VAT rating from the DMS submission payload

Dweik97:


glidester (changes requested):
Can we add a unit test for this logic pls

PR #544, opened 2024-05-22
CEDS-5824 Optimize the payload of draft Declarations

udinobi:


app/uk/gov/hmrc/exports/services/DeclarationService.scala
```diff
@@ -73,15 +71,37 @@ class DeclarationService @Inject() (declarationRepository: DeclarationRepository
...
+    isAmendment: Boolean
+  ): Future[Some[(Boolean, String)]] = {
+    val associatedSubmissionId = if (isAmendment) declaration.declarationMeta.associatedSubmissionId else None
+    val declarationMeta = declaration.declarationMeta
+      .copy(
+        parentDeclarationId = Some(parentId),
+        parentDeclarationEnhancedStatus = Some(enhancedStatus),
+        status = if (isAmendment) AMENDMENT_DRAFT else DRAFT,
+        associatedSubmissionId = associatedSubmissionId
+      )
```
glidester, line 87: Not related to this ticket, but as you are touching this section of code....
Can you set the `createdDateTime` and `updatedDateTime` fields to equal `now` when we make this copy please.
We ran across this issue yesterday when playing with the saved declaration's screen, when making a new copy of a declaration it should not retain the original's timestamp values. Thanks :slightly_smiling_face:

conf/app.routes
```diff
@@ -1,6 +1,6 @@
 # JSON Based
 POST        /declarations                                           uk.gov.hmrc.exports.controllers.DeclarationController.create
-GET         /declarations                                           uk.gov.hmrc.exports.controllers.DeclarationController.findAll(status: Seq[String] ?= Seq.empty, page: Page ?= Page(), sort: DeclarationSort ?= DeclarationSort())
+GET         /declarations                                           uk.gov.hmrc.exports.controllers.DeclarationController.fetchPage(status: Seq[String] ?= Seq.empty, page: Page ?= Page(), sort: DeclarationSort ?= DeclarationSort())
```
glidester, line 3: Would it not be safer to define a new endpoint url that has the new behavior and leave the existing endpoint as is (to be cleaned up later)?
This way we can deploy the BE first with no impact on the FE, then deploy the FE afterwards. At the moment which ever we deploy first FE or BE will cause the service to error (especially as we now have 14 instants to deploy to in prod!).

glidester (changes requested):

glidester (approved):

PR #553, opened 2024-06-28
CEDS-5912 Remove non-compliant characters from the XML payload before submitting a declaration

udinobi:


app/uk/gov/hmrc/exports/controllers/SubmissionController.scala
```diff
@@ -40,14 +41,23 @@ class SubmissionController @Inject() (
...
   def create(declarationId: String): Action[AnyContent] = authenticator.authorisedAction(parse.default) { implicit request =>
-    declarationService.markCompleted(request.eori, declarationId, declarationId).flatMap {
+    def submitDeclaration(declaration: ExportsDeclaration): Future[Submission] =
+      try
+        submissionService.submit(declaration)
+      catch {
+        case throwable: Throwable =>
+          logger.info(s"Declaration [${declaration.id}]: Submission failed")
+          declarationService.revertStatusToDraft(declaration, throwable)
+      }
```
glidester, line 51: I think this reverting on exception should be the responsibility of the SubmissionService class rather than this controller.

Imagine if we wanted to call the `submit` method from another controller as well, then that call would be missing this protection.

glidester (changes requested):

app/uk/gov/hmrc/exports/services/mapping/Sanitiser.scala
```diff
@@ -0,0 +1,87 @@
...
+      case '\"' => if (isAttVal) writer.write("&quot;") else writer.write('\"')
+      case '&'  => writer.write("&amp;")
+      case '<'  => writer.write("&lt;")
+      case '>'  => writer.write("&gt;")
+      case _ =>
+        if ((chr > 31 && chr < 127) || isLetterOrDigit(chr)) writer.write(chr.toInt)
+        else if (isWhitespace(chr) || isISOControl(chr)) writer.write(' ')
+        else writer.write(s"&#${chr.toInt};")
+    }
+}
```
glidester, line 72: Why are we duplicating the logic here?

udinobi: Check the `if` condition at line #37. This method handles buffers with a single char. This improves the performance because we do not need to create a new `StringBuilder` for every value

glidester: Yes I see that, just thinking to avoid duplicating the case matching logic, maybe something like this:

udinobi: Honestly... you try to avoid the duplication (and I'm conscious of that but IMO I do not think it is really important here) whereas I'm more concerned about the performance of the conversion process.

Please consider these improvements over your version:

- I do not create a new String for values consisting of a single char (and we have several of those in a declaration)
- `writer.write(Char)` is more performant than `writer.write(String)`

Also note that after last commit a few `case`s have been removed, so lowering the duplication.


Just my 2 cents :slightly_smiling_face:

app/uk/gov/hmrc/exports/controllers/SubmissionController.scala
```diff
@@ -41,13 +41,13 @@ class SubmissionController @Inject() (
...
-        if (declarationBeforeUpdate.isCompleted) {
-          Future.successful(Conflict(ErrorResponse("Declaration has already been submitted")))
-        } else {
-          submissionService.submit(declarationBeforeUpdate).map(Created(_))
-        }
+        if (declarationBeforeUpdate.isCompleted) Future.successful(Conflict(ErrorResponse("Declaration has already been submitted")))
+        else
+          for {
+            submissionToSave <- submissionService.submitDeclaration(declarationBeforeUpdate)
+            submissionToReturn <- submissionService.storeSubmission(submissionToSave)
```
glidester, line 49: Should this action also not be brought under the responsibility of the SubmissionService rather than the controller? :thinking:

glidester (changes requested):

app/uk/gov/hmrc/exports/controllers/SubmissionController.scala
udinobi: Oh, yes, yes. Well spotted. Forgot to do that previously.

glidester (approved):

PR #564, opened 2024-09-20
CEDS-6059 Set 'ReattemptNotificationParsingRoutine' as a scheduled job and remove the 'routines' package

udinobi:


app/uk/gov/hmrc/exports/scheduler/jobs/PurgeAncientSubmissionsJob.scala
```diff
@@ -35,23 +35,18 @@ class PurgeAncientSubmissionsJob @Inject() (
...
 
-  override val name: String = "PurgeAncientSubmissions"
-
-  override def interval: FiniteDuration = appConfig.purgeAncientSubmissions.interval
-
   override def firstRunTime: Option[LocalTime] = Some(appConfig.purgeAncientSubmissions.elapseTime)
+  override def interval: FiniteDuration = appConfig.purgeAncientSubmissions.interval
 
-  lazy val expiryDate = {
+  lazy val expiryDate: ZonedDateTime = {
```
glidester, line 44: As this class is a singleton, and this date is a val, then this date will only ever change when the application is redeployed. So that means if we don't redeploy for a week then submissions will start to build up that are older than the 180 days! I think this date (and the filter) need to be methods rather than vals :thinking:

